PR #61, opened 2018-01-23
Updated README Links

aemc:


ZirconiumX:
I've opened #62, which is this branch with the correct merge target.

PR #65, opened 2018-01-24
Replace qmake with CMake, ckb with ckb-next

light2yellow:
Rationale
=========

CMake:
  * possibility to work on the project using IDEs and editors with CMake integration
  * lack of extensibility and features by qmake
  * macOS toolchain bugs of qmake that prevented patch-free compilation
  * CMake in 2017 has become a standard when speaking of build systems

ckb -> ckb-next backwards-incompatible step:
  * since the beginning of ckb-next, the `-next` transition was concurrent with the development of listfiles, hence the inclusion of ckb -> ckb-next transition into the buildsystem, in order to not redo already done job, since the direct relation of buildsystem to the targets

Layout:
  * unsatisfactory one, with irrational paths and poor logic in names

Summary
=========

Project layout has changed, see 26bafca.

Executables were renamed to reflect the transition away from ccMSC's ckb.

`ckbnextconfig.h` is used to create the layer of communication between the buildsystem and the built software.

All authored files are distributed under BSD 3-clause. The authorship is passed to a group of people who are the members of ckb-next GitHub organisation[0] at the time of writing: fleischie, frickler24, light2yellow, tatokis, ZirconiumX.

Description of installation process, troubleshooting etc. is removed from the README and moved into the GitHub Wiki. See Home page[1] for the list of currently available pages.

Effort has been put into the general improvement of compilation, installation and running experience pipeline. The amount of obscure, system-configuration-dependent issues is expected to be lower significantly due to determinism and reproducibility in mind while creating the ecosystem, which is the first step to achieve same in the built software itself.

[0]: https://github.com/ckb-next/people
[1]: https://github.com/ckb-next/ckb-next/wiki


light2yellow:
FYI, if you decide to merge: should be merged with the commit message filled with the above exactly as written (press 'Edit'), hard-wrapped at 72 on word boundary.

src/animations/heat/main.c
```diff
@@ -1,28 +1,27 @@
...
 #include <math.h>
+#include <ckb-next/animation.h>
 
 void ckb_info(){
     // Plugin info
     CKB_NAME("Heat Map");
     CKB_VERSION("0.2");
     CKB_COPYRIGHT("2017", "RULER501");
-    CKB_LICENSE("GPLv2");
+    CKB_LICENSE("GPLv2+");
```
tatokis: I don't think we're allowed to do this...

src/animations/mviz/main.c
```diff
@@ -1,35 +1,32 @@
...
+#include <kissfft/kiss_fftr.h>
+#include <ckb-next/animation.h>
 
 void ckb_info(){
     // Plugin info
     CKB_NAME("Music Visualization");
     CKB_VERSION("0.2");
     CKB_COPYRIGHT("2017", "RULER501");
-    CKB_LICENSE("GPLv2");
+    CKB_LICENSE("GPLv2+");
```
tatokis: Likewise

src/gui/media_mac.m
```diff
@@ -1,5 +1,7 @@
-#import <AudioToolbox/AudioServices.h>
-#import <Foundation/Foundation.h>
+#include <AudioToolbox/AudioServices.h>
+#include <Foundation/Foundation.h>
+//#include <CoreServices/CoreServices.h>
+#include <CoreAudio/CoreAudio.h>
```
tatokis: Why is this change in cmake?

src/gui/resources/demoprofile.conf
```diff
@@ -629,4 +629,4 @@ CurrentMode={11C11AE3-3195-4DFC-B8AC-2FEA703414E5}
 HwModified=40b6f054
 ModeCount=3
 Modified=2015a5b9
-Name=ckb Demo
+Name=ckb-next Demo
```
tatokis: I think it might be better if we simplify this to just "Demo".

src/gui/animscript.cpp
```diff
@@ -27,11 +28,7 @@ AnimScript::~AnimScript(){
 }
 
 QString AnimScript::path(){
-#ifdef __APPLE__
-    return QDir(QApplication::applicationDirPath() + "/../Resources").absoluteFilePath("ckb-animations");
-#else
-    return QDir("/usr/lib").absoluteFilePath("ckb-animations");
-#endif
+    return QString(CKB_NEXT_ANIMATIONS_PATH);
```
tatokis: Not strictly related to this, but maybe in the future we should return a QStringList or something along those lines, so that we can have the possibility of both user-defined and local-relative paths for animations.

src/animations/heat/main.c
light2yellow: According to https://github.com/ckb-next/ckb-next/commit/6b8c2b7fc97244b6fee7534953622bd0ae86ecb5 it is provided under GPLv2+, so I just fixed it in the macro. I should do this on master, though.

src/animations/mviz/main.c
light2yellow: According to https://github.com/ckb-next/ckb-next/commit/9704df68efbe67ab0303eda184999521647a1b3e it is provided under GPLv2+, so I just fixed it in the macro. I should do this on master, though.

src/gui/media_mac.m
light2yellow: It builds fine without it with qmake, but the build fails without these headers with CMake. I never figured out, why. I will try again without it in the nearest future, because at the time I wrote this I wasn't totally sure of what I was doing, now I am. However, I assume qmake tries to be smart, it never required CoreAudio. It should be rebased though, as `include` is the same as `import` in ObjC (but who knows) and I was desperate to make it built (since whatever I did failed), so I tried different combinations until it worked. The diff is basically `+CoreAudio` here. Why is it the way it is is still a mystery to me.

src/gui/resources/demoprofile.conf
light2yellow: Ok.

src/gui/animscript.cpp
light2yellow: Ok, leaving it to be done afterwards, after merge.

src/gui/media_mac.m
light2yellow: I guess I recall: I see there were 10.12 API changes (deprecation) which the API has warned me about, and which I tried to fix. I am not sure if during qmake these warnings have appeared, though. Need to check on Hackintosh, although it's 10.13.x and I remember that 10.12 introduced some deprecation warnings which did not provide any replacement, so it's all complicated and might not be reproducible, since they might have fixed it in 10.13.

light2yellow:
Comments were addressed, branch was tested on macOS and another round of issue-fixing has occurred, not related to the comments. It is ready for another review.

light2yellow:
@tatokis 
Can't reproduce - it compiles. Provide the full output of `./quickinstall`, or whatever. Also make sure your local branch is up to date with the remote.

tatokis:
@light2yellow https://vps.tasossah.com/bugs/ckb-next/ckb_cmake_error.txt

light2yellow:
@tatokis 
Can't figure anything out of this, because it does generate the ui headers in your log:
```
Generating ui_kbbindwidget.h
[ 30%] Building C object src/daemon/CMakeFiles/ckb-next-daemon.dir/input.c.o
Generating ui_kblightwidget.h
Generating ui_kbprofiledialog.h
[ 31%] Building C object src/daemon/CMakeFiles/ckb-next-daemon.dir/keymap.c.o
Generating ui_kbwidget.h
Generating ui_kblightwidget.h
Generating ui_kperfwidget.h
Generating ui_layoutdialog.h
Generating ui_mainwindow.h
Generating ui_modeselectdialog.h
Generating ui_mperfwidget.h
Generating ui_rebindwidget.h
```
So, this is confusing.

Please post the output of:
```
export LC_ALL=C
cd <project source dir>
git clean -xdf
git status
cmake -H. -Bbuild
cmake --build build --target ckb-next_autogen -- VERBOSE=1
cmake --build build --target ckb-next -- VERBOSE=1
```

light2yellow:
Rebased to apply onto the latest changes from master, nothing cmake-related has changed.

quickinstall
```diff
@@ -1,249 +1,45 @@
...
-            finish 1
-        fi
-    fi
-    # Check for Xcode presence
-    if xcodebuild -version 2>&1 | grep "error"; then
-        echo "Please install Xcode and rerun the script."
-        finish 1
-    fi
-fi
+git clean -xdf
```
fleischie: Super small concern: What if the user downloaded the source code via zip?

I admit it's a super obscure edge-case, but this means we should have `git` as a dependency for this project.

tatokis: I don't think calling git clean in quickinstall is a good idea.
People tend to make changes we tell them to do, and then call quickinstall, as they don't know how else to compile it, before we even have a chance to say anything.

tatokis:
@light2yellow The commands you sent me do not work, and also ran git clean like an idiot which deleted all of my local patches.
```
tatokis@tatokis-PC:~/ckb-next/newrepo/ckb-next$ cmake --build build --target ckb-next_autogen -- VERBOSE=1
make: *** No rule to make target 'ckb-next_autogen'.  Stop.
```

Anyway, I ran with VERBOSE=1, the headers are generated by uic, and I can verify their existence.
```
Generating ui_animadddialog.h
/usr/lib/x86_64-linux-gnu/qt5/bin/uic -o /home/tatokis/ckb-next/newrepo/ckb-next/build/src/gui/ui_animadddialog.h /home/tatokis/ckb-next/newrepo/ckb-next/src/gui/animadddialog.ui 
```
However, after running `strace -f -e open make`, I can see that the compiler looks for the in the wrong directories.
```
[pid 15579] open("/home/tatokis/ckb-next/newrepo/ckb-next/src/gui/ui_animadddialog.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
[pid 15579] open("/home/tatokis/ckb-next/newrepo/ckb-next/build/ui_animadddialog.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
```

light2yellow:
@tatokis's build problem is solved.

light2yellow:
@fleischie no more `BUILD`, no more `git clean`.

tatokis:
@light2yellow Regarding versioning, it would be nice if the `project(ckb-next LANGUAGES C CXX VERSION 0.2.9)` line was moved to something like VERSION.cmake so that it can easily be updated on release, and also introduce a flag (or reuse an existing one), that would have two states.

ON would mean release build, which will be set for non-git zip/tar _release_ archives.
OFF would mean non-release zip build, which will be the default value in the repo.
The OFF flag should set a -zip suffix to the version string, so that such builds do not get confused with actual release builds.

To sum up, on every release, a commit will be made, incrementing the hardcoded version string, a tag will also be set.
The zip archives will have the above flag set to ON, so that nothing is appended to the version string.
During configuration, cmake will:
a. check if git is available, and if so extract the version from there.
b. if git is not available, read the hardcoded version and check the state of the flag.
If the flag is set to ON, that means a zip/tar release is being compiled, so the version is set to the hardcoded one and no suffix is added.
If the flag is set to OFF, that means it's a zip copy from master, so the version is set to the hardcoded one, and the suffix -zip is added.

tatokis:
If no one has any objection, I am ready to merge this.

tatokis:
I take that back. Not only does the GUI not link against libpulse, the cmake module to detect it doesn't use pkg-config, but instead searches through specified paths.

I'll have to rewrite that, link the GUI against pulse, and set the appropriate define for it.

ZirconiumX:
I think libpulse shouldn't block this, considering our recent issues with it, and that the GUI does okay without it.

PR #68, opened 2018-01-30
Specify support for the K68 non-RGB version only

mauriciofauth:
Corsair has announced the RGB version of the K68.


ZirconiumX:
Would you mind doing a `sed -i 's/K68/K68_NRGB/' src/ckb-daemon/*`?

mauriciofauth:
@ZirconiumX no problem, done in #71

PR #70, opened 2018-01-30
Add openSUSE preparation entry

BenWhitehead:


light2yellow:
@BenWhitehead I am working on #65, which is responsible for https://github.com/ckb-next/ckb-next/wiki/Linux-Installation and therefore need a few additional details - in particular, packages for `cmake` (at least 3.5+), `libappindicator` (Ayatana lib, either gtk2 or gtk3 backend, doesn't matter), `libpulse` (pulseaudio lib), `quazip` (built against qt5). I would be glad if you can help me, since when I try to google for it I get `Error: OBS Backend not available`, which is sad to see considering SUSE actually earns money and should be able to keep their infrastructure at least running.

BenWhitehead:
@light2yellow Here is a summary of those packages I was able to figure out

| package | devel-pkg | version |
|--- | --- | --- |
| `cmake` | | 3.10.2-1.1 |
| `libappindicator1` | `libappindicator-devel` | 12.10.1+bzr20170215-2.4 |
| `libpulse0` | `libpulse-devel` | 11.1-2.1 |
| `libquazip5-1` | `quazip-qt5-devel` | 0.7.3-1.1 |


Hope this helps.

light2yellow:
Thanks, appreciated. Both Fedora and openSUSE web interfaces to represent packages are awful, I would burn them with fire.

PR #79, opened 2018-02-07
Polaris mousepad support

ZirconiumX:
See #55 and mattanger/ckb-next#381.


ZirconiumX:
Note that we plan to make an mm800-redux branch which has this all neatly configured, but first we need to make sure this works.

ZirconiumX:
We're just waiting on the interrupt-read-fix branch, and then I'll PR mm800-redux. Until then...

PR #83, opened 2018-02-09
Skip BEGIN FW ENTRIES in the FW table

tatokis:
Minor edit to prevent the GUI from throwing a warning when reading the firmware table.


light2yellow:
+1 to @ZirconiumX, I think we can resolve small things post-factum if anything bad happens. Not a big deal.

Issue #82, opened 2018-02-09
Invisible mouse wall when gaming with CKB


Thalagyrt:
Hiya! I apologize for not having more diagnostic information as I have uninstalled the CKB driver entirely - if you need it, I will reinstall it temporarily to get it. I was using v0.2.9 from the releases pages with both a Corsair Scimitar and Corsair M65 and running into issues while gaming.

When playing a game that requires click + drag or right click + drag as in many shooters, the OSX cursor moves, hidden, while running the CKB driver. Eventually it hits the edge of the screen and cannot move any further, and the visual result is that in game you hit an invisible wall where you can't turn anymore. For example, while playing Skyrim via Steam Streaming, I would try shooting a bow only to find that I can't track my target because the cursor is hitting an invisible wall past which it cannot move. This only happens with a button pressed down, and happens in Mac native games, happens in Geforce Now, and also in Steam Streaming.

With CKB uninstalled and using the native OSX mouse drivers, the problem disappears entirely and everything behaves as expected. The problem occurs whenever ckb-daemon is running, whether or not CKB the app is running or not, which initially led me to believe it was an OSX problem since I thought I had completely exited CKB. When I found nobody else on any forums with the same issue, I suspected the mouse driver and looked into how to completely remove it and the problem went away completely.

If there's any way I can reinstall temporarily and get you some logs/extra info I'd be happy to! If you want an easy target to test this with on your own machine, Golf With Your Friends is $6 on Steam and is Mac native and exhibits the problem with CKB as you have to click and drag forward to make a shot. The invisible wall effect leads to you being unable to shoot at full power in many situations.

Thanks so much for all the awesome work on this otherwise phenomenal driver, and I look forward to being able to use it again!

ZirconiumX:
I distinctly remember somebody else posting about this bug, I'll see if I can find it.

Thalagyrt:
Wow, that was a fast reply! I'm happy to do as much testing as you want. Hopefully we can find a fix. :)

Thalagyrt:
Also if you don't have any games to test with, I'll happily gift you one of the ones that had the problem on Steam so you can test as I don't feel you should have to pay out of pocket to fix this. You do enough as is maintaining it.

ZirconiumX:
Actually, the problem is a bit more serious than a game that costs a couple of bucks; only @fleischie has a Mac - and a job - and @light2yellow has a Hackintosh, so we're a bit limited in terms of hardware and reproducibility.

[...]

PR #85, opened 2018-02-09
Add conditional on OSX pre-sierra hack to not run in Sierra+

Thalagyrt:
- Fetch OSX version from kern.osrelease sysctl
- All credit for finding the source of the bug goes to @ofBits.
  https://github.com/ckb-next/ckb-next/issues/28#issuecomment-359292651

I apologize if this is ugly hacky garbage and welcome feedback on a better way to split the string - I haven't touched C in over a decade, and I'm well aware this won't work once we hit version 100 of OSX. I think that's a few decades off though, given we're at 17 now. :)

I've tested this in games on High Sierra and confirm that the mouse hitting an invisible wall is gone. I do not have a pre-Sierra machine to test that the old hack still runs there. I do not see any noticeable difference in performance from the additional branches - the mouse is still buttery smooth and quickly responsive.

I am not sure if the `errno.h` include is necessary to use the sysctl - to be honest that was just a StackOverflow grab. If it's unnecessary I probably should remove it!

For more information, see discussion in #23, #28, and #82

### DIFF
--- a/src/ckb-daemon/input_mac.c
+++ b/src/ckb-daemon/input_mac.c
@@ -3,6 +3,11 @@
 #include "input.h"
 
 #ifdef OS_MAC
+#include <errno.h>
+#include <sys/sysctl.h>
+
+bool osx_sierra;
+bool checked_for_osx_sierra = false;
 
 // Numpad keys have an extra flag
 #define IS_NUMPAD(scancode) ((scancode) >= kVK_ANSI_KeypadDecimal && (scancode) <= kVK_ANSI_Keypad9 && (scancode) != kVK_ANSI_KeypadClear && (scancode) != kVK_ANSI_KeypadEnter)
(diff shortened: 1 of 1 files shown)


tatokis:
@ZirconiumX I doubt anyone is going to see this and test it, just like no one did for the modifiers for like half a year. I'd say just merge it (and watch people complain).

Thalagyrt:
Big thanks to both of you for the quick engagement, advice, and helping me get this together. If there's a way I can buy you both some beers let me know. :)

ZirconiumX:
Unfortunately, @tatokis doesn't drink.

Thalagyrt:
Ah, totally respect that!

Thalagyrt:
A thought occurred to me that on OSX 10.7 (or 10.6, whichever is technically version 9 or earlier) we'll call `atoi("9.")`. Thankfully this results in 9, so we should be OK - just in case anyone else has the same thought I suddenly had. I figure you probably didn't worry about it because you all know C way better than me.

src/ckb-daemon/input_mac.c
```diff
@@ -25,15 +30,29 @@ static void postevent(io_connect_t event, UInt32 type, NXEventData* ev, IOOption
...
-        location.y = floor(loc.y + ev->mouseMove.dy);
-        options = (options & ~kIOHIDSetRelativeCursorPosition) | kIOHIDSetCursorPosition;
+
+    if(!checked_for_osx_sierra){
+        char osx_version_buf[256];
+        size_t size = sizeof(osx_version_buf);
+        sysctlbyname("kern.osrelease", osx_version_buf, &size, NULL, 0);
+        char osx_version[2];
+        strncpy(osx_version, osx_version_buf, 2);
+        osx_sierra = atoi(osx_version) >= 16;
```
light2yellow: This is wrong. For example, `10.12.1` is built on XNU `15.6.0`. Moreover, trying to detect OS based on the kernel version is wrong - they don't depend on each other. I think we will have to find a better way. I tested different constants and variables exposed by `CTL_KERN`, to no avail. And that's reasonable, since kernel can't know about macOS version (which is done in a BSD style, not in a Linux style).

light2yellow: I am referring to https://github.com/dylanaraps/neofetch/issues/433. Even though [Wikipedia](https://en.wikipedia.org/wiki/Darwin_%28operating_system%29#Release_history) says it should be 16.0, so Wikipedia must be wrong, too.

Thalagyrt: Very good point. I'll do some more digging over the weekend and try to find a better way.

Thalagyrt: I do know that we can shell out to `sw_vers -productVersion` to get the version number, but that's hacky and there's got to be a better way.

light2yellow: TLDR; this is correct, sorry.

So, I took another approach. I went to opensource.apple.com, chose macOS 10.12 and looked at the XNU package version, which is `3789.1.32`. Great, thought I, this should be _the_ way to check for the kernel, since I would know exactly which kernel build is supplied with 10.12.0. I downloaded it to determine how this version is even generated and what does it have to do with the normal version. The answer is in `libkern/libkern/version.h.template`:

```c
/* version is a string of the following form, as returned by uname -v:
 * "Darwin Kernel Version <osrelease>: <build date>; <osbuilder>:<build root>"
 */
extern const char version[];
```

A `ripgrep "version\["` revealed the existence of `config/version.c`, which contains a comment:
```c
/* version.c
 * This file is a C source template for version.c, which is generated
 * on every make of xnu.  This template is processed by the script
 * xnu/config/newvers.pl based on the version information in the file
 * xnu/config/MasterVersion.
 */
```

which in turn led to `config/MasterVersion`:
```pl
16.0.0

# The first line of this file contains the master version number for the kernel.
# All other instances of the kernel version in xnu are derived from this file.
#
# The format of the version number must conform to the version resource format
# as described in TN1132: http://developer.apple.com/technotes/tn/tn1132.html
#
# In particular, the string is formatted as: J[.N[.R[S[L]]]], where:
#  J represents the kernel major version number (integer)
#  N represents the kernel minor version number (integer)
#  R represents the kernel revision number (integer)
#  S represents the kernel build stage (one of "d", "a", "b", or "r")
#  L represents the kernel pre-release level (integer)
#
# The correct way to make use of the kernel version within kernel code or a
# kext is to include libkern/verison.h.  version.h contains defines that can
# be used for build-time version logic and prototypes for variables that can
# be used for run-time version logic.
```

So, this confirms that `16` is the correct major version. I also downloaded `xnu-3248.60.10` supplied with `10.11.6` and `config/MasterVersion` contained `15.6.0`.

Funnily enough, if you try to actually follow the technical note link, you'll find nothing. Fortunately, there's a [mirror](http://mirror.informatimago.com/next/developer.apple.com/technotes/tn/tn1132.html) of it. Apple's versioning system is an entirely different animal.

Thalagyrt:
Thanks for digging deeper, and again thanks to you all for being so welcoming and helping get this fix out! You all rock!

ZirconiumX:
"Don't be a dick" is one of our two rules, and I try to take it to heart.

Thalagyrt:
"Don't be a dick" is verbatim the only major rule of every online community I run.

PR #87, opened 2018-02-11
Colourpicker warning

tatokis:
Add a warning that gets displayed if QT_QPA_PLATFORMTHEME is neither kde nor qt5ct.
If the colour picker doesn't fail to draw, the notification is marked as dismissed and never displayed again.


light2yellow:
Fixed up and merged as 4e0ae2015302057a528b22a14879d6ab5a41f742.

ZirconiumX:
It's still a workaround if you're working around the users, right?

PR #88, opened 2018-02-11
Read wMaxPacketSize from udev

tatokis:
Due to some devices breaking when assuming MSG_SIZE for max buffer length, we now read wMaxPacketSize from the descriptor using udev.


tatokis:
@ZirconiumX thanks for pointing it out, make debug doesn't seem to throw any warnings.
I'll fix it before getting this merged.

light2yellow:
Fixed up, rebased and merged as 0db73f4edaec5cda39e1174091cbdc24bb19ec29.

PR #98, opened 2018-02-16
Use URB Interrupts for all non-HID transfers

### LINKED ISSUE #51: [Long-term] Switch to URB Interrupts for all transfers
<a href="https://github.com/ZirconiumX"></img></a> **Issue by [ZirconiumX](https://github.com/ZirconiumX)**
_Tuesday Dec 19, 2017 at 10:20 GMT_
_Originally opened as https://github.com/mattanger/ckb-next/issues/367_

----

## Boring technical stuff, TL;DR below
I'll begin with some context for those of you not on the IRC (@frickler24 comes to mind):

USB has four kinds of data transfer:
- Control, used mostly for setup and initialisation.
- Bulk, used for USB sticks and mass data transfer.
- Interrupt, used for keyboards and mice, mostly.
- Isochronous, used for USB sound cards.

Side note: to send a URB Interrupt through the Linux `ioctl` interfaces, you need to perform a single packet URB Bulk transfer. Don't ask why, I don't know either.

At the moment, for the Linux USB backend at least, we [send](https://github.com/ckb-next/ckb-next/blob/master/src/ckb-daemon/usb_linux.c#L70-L77) URB Interrupts for write packets, but [send](https://github.com/ckb-next/ckb-next/blob/master/src/ckb-daemon/usb_linux.c#L79-L80) and [receive](https://github.com/ckb-next/ckb-next/blob/master/src/ckb-daemon/usb_linux.c#L141-L142) URB Controls for read packets. Meanwhile, an input thread [receives](https://github.com/ckb-next/ckb-next/blob/master/src/ckb-daemon/usb_linux.c#L298-L311) URB Interrupts for input packets.

This worked fine for version 1 and 2 of the Corsair firmwares (excepting the [K95 Platinum](https://github.com/ckb-next/ckb-next/issues/53#issuecomment-290994546)), but it seems a portion of version 3 firmwares (the K95P again, my M65 Pro) do not support Control transfers at all in software control mode. This means we need to switch to Interrupts for all packets.

Those of you paying attention may have noticed [the commented section in os_usbrecv](https://github.com/ckb-next/ckb-next/blob/master/src/ckb-daemon/usb_linux.c#L132-L140), in particular, `This is what CUE does, but it doesn't seem to work on linux`.

On Saturday, I did some digging into why it doesn't work, and found that the reply packets were being received by the input thread, so the main thread that wished to receive an event would never get the packet.

This is an awkward problem, because it means we have to transfer data between threads, which is non-trivial.

## If you're not a techie, here's the TL;DR:
On firmware version 3 and the K95P, you won't be able to load stored settings on your device. For now we suggest you don't upgrade the firmware version, and bear with us.

tatokis:
Closes #51, which contains a really nice summary of the whole problem.

A mutex and a buffer are added to the usbdevice struct.

When usbsend is called with the is_recv flag, the mutex is locked.
The main thread listens for packets starting with 0x0e, copies them to the buffer and unlocks the mutex.
usbrecv waits for the mutex to be unlocked and then copies the buffer from usbdevice.

The above is all ignored, falling back to the old behaviour, when fwversion is lower than 1.3 which works only with URB Controls.

Ideally, the comments for the functions should be rewritten before merging this.

### DIFF
--- a/src/ckb-daemon/usb_linux.c
+++ b/src/ckb-daemon/usb_linux.c
@@ -8,9 +8,19 @@
 
 /// \details
 /// \brief all open usb devices have their system path names here in this array.
-#define DEBUG
 
 static char kbsyspath[DEV_MAX][FILENAME_MAX];
+
+/// Formats and writes the current urb buffer to the console
+void print_urb_buffer(const char* prefix, const unsigned char* buffer, int actual_length, const char* file, int line, const char* function){
+    char converted[actual_length * 3 + 1];
+    for(int i = 0; i < actual_length; i++)
+        sprintf(&converted[i * 3], "%02x ", buffer[i]);
+    if(line == 0)
+        ckb_info("%s %s\n", prefix, converted);
+    else
+        ckb_info("%s (via %s:%d) %s %s\n", function, file, line, prefix, converted);
+}
 
 ////
 /// \brief os_usbsend sends a data packet (MSG_SIZE = 64) Bytes long
@@ -67,7 +77,12 @@ static char kbsyspath[DEV_MAX][FILENAME_MAX];
 ///
 int os_usbsend(usbdevice* kb, const uchar* out_msg, int is_recv, const char* file, int line) {
     int res;
-    if ((kb->fwversion >= 0x120 || IS_V2_OVERRIDE(kb)) && !is_recv){
+
+    if (kb->fwversion >= 0x120 || IS_V2_OVERRIDE(kb)){
+        // If we need to read a response, lock the mutex so that os_usbrecv() waits for the main thread.
+        if(is_recv)
+            if(!!pthread_mutex_lock(&kb->interruptmutex))
+                ckb_fatal("Error locking interrupt mutex in os_usbsend()\n");
         struct usbdevfs_bulktransfer transfer = {0};
         // All firmware versions for normal HID devices have the OUT endpoint at the end.
         // Devices with no input, such as the Polaris, have it at the start.
@@ -83,19 +98,18 @@ int os_usbsend(usbdevice* kb, const uchar* out_msg, int is_recv, const char* fil
     }
 
     if (res <= 0){
-        ckb_err_fn(" %s, res = 0x%x\n", file, line, res ? strerror(errno) : "No data written", res);
-        if(res == -1 && errno == ETIMEDOUT)
+        int ioctlerrno = errno;
+        ckb_err_fn(" %s, res = 0x%x\n", file, line, res ? strerror(ioctlerrno) : "No data written", res);
+        if(res == -1 && ioctlerrno == ETIMEDOUT)
             return -1;
         else
             return 0;
     } else if (res != MSG_SIZE)
         ckb_warn_fn("Wrote %d bytes (expected %d)\n", file, line, res, MSG_SIZE);
 #ifdef DEBUG_USB_SEND
-    char converted[MSG_SIZE*3 + 1];
-    for(int i=0;i<MSG_SIZE;i++)
-        sprintf(&converted[i*3], "%02x ", out_msg[i]);
-    ckb_warn_fn("Sent %s\n", file, line, converted);
+    print_urb_buffer("Sent:", out_msg, MSG_SIZE, file, line, __func__);
 #endif
+
     return res;
 }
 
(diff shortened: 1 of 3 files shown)


src/ckb-daemon/usb_linux.c
```diff
@@ -3,6 +3,7 @@
 #include "input.h"
 #include "notify.h"
 #include "usb.h"
+#include "time.h"
```
tatokis: This probably shouldn't be here...

light2yellow: With CMake, I can implement an include-what-you-use runner in future.

src/ckb-daemon/usb_linux.c
```diff
@@ -131,32 +136,37 @@ int os_usbsend(usbdevice* kb, const uchar* out_msg, int is_recv, const char* fil
...
-        ckb_err_fn("%s\n", file, line, res ? strerror(errno) : "No data read");
-        if(res == -1 && errno == ETIMEDOUT)
-            return -1;
-        else
-            return 0;
-    } else if(res != MSG_SIZE)
-        ckb_warn_fn("Read %d bytes (expected %d)\n", file, line, res, MSG_SIZE);
+        if(res <= 0){
+            ckb_err_fn("%s\n", file, line, res ? strerror(errno) : "No data read");
+            if(res == -1 && errno == ETIMEDOUT)
```
light2yellow: `ckb_err_fn()` calls `fprintf()` which is allowed to set `errno`, so this branch is not taken if `fprintf` encounters an error.

src/ckb-daemon/usb_linux.c
```diff
@@ -319,6 +329,10 @@ void* os_inputmain(void* context){
         ioctl(fd, USBDEVFS_SUBMITURB, urbs + i);
     }
 
+    /// Allocate memory for the os_usbrecv() buffer and create the mutex
+    kb->interruptbuf = malloc(MSG_SIZE * sizeof(uchar));
+    pthread_mutex_init(&kb->interruptmutex, NULL);
```
light2yellow: Missing error checking and handling, missing `pthread_mutex_destroy`.

src/ckb-daemon/usb_linux.c
```diff
@@ -357,46 +371,61 @@ void* os_inputmain(void* context){
...
-                    break;
-                case MSG_SIZE:
-                    // RGB EP 3: Corsair input
-                    corsair_kbcopy(kb->input.keys, -urbendpoint, urb->buffer);
-                    break;
-                }
+
+            #ifdef DEBUG_USB_INPUT
+            char converted[urb->actual_length*3 + 1];
+            for(int i=0;i<urb->actual_length;i++)
```
light2yellow: Please format for readability.

tatokis: This was a direct copy from https://github.com/ckb-next/ckb-next/blob/master/src/ckb-daemon/usb_linux.c#L154-L159

src/ckb-daemon/usb_linux.c
tatokis: Thought the latter was added, will correct.
Although short of displaying a message, not sure what kind of error handling can be done in this case.

light2yellow: Since the attributes on init() are NULL, instead of init() and destroy() consider using PTHREAD_MUTEX_INITIALIZER.

src/ckb-daemon/usb_linux.c
```diff
@@ -319,6 +329,10 @@ void* os_inputmain(void* context){
         ioctl(fd, USBDEVFS_SUBMITURB, urbs + i);
     }
 
+    /// Allocate memory for the os_usbrecv() buffer and create the mutex
+    kb->interruptbuf = malloc(MSG_SIZE * sizeof(uchar));
```
light2yellow: Error checking is missing (malloc return value).

src/ckb-daemon/usb_linux.c
```diff
@@ -357,46 +371,61 @@ void* os_inputmain(void* context){
...
-                case MSG_SIZE:
-                    // RGB EP 3: Corsair input
-                    corsair_kbcopy(kb->input.keys, -urbendpoint, urb->buffer);
-                    break;
-                }
+
+            #ifdef DEBUG_USB_INPUT
+            char converted[urb->actual_length*3 + 1];
+            for(int i=0;i<urb->actual_length;i++)
+                sprintf(&converted[i*3], "%02x ", ((uchar*)urb->buffer)[i]);
```
light2yellow: Consider using snprintf instead. Consider checking for errors.

src/ckb-daemon/usb_linux.c
```diff
@@ -357,46 +371,61 @@ void* os_inputmain(void* context){
...
+            ckb_info("Input Recv %s\n", converted);
+            #endif
+
+            // If the response starts with 0x0e, that means it needs to go to os_usbrecv()
+            if(urb->actual_length == MSG_SIZE && ((uchar*)urb->buffer)[0] == 0x0e){
+                memcpy(kb->interruptbuf, urb->buffer, MSG_SIZE);
+                // Unlock the mutex, signaling os_usbrecv() that the data is ready.
+                pthread_mutex_unlock(&kb->interruptmutex);
```
light2yellow: Consider checking for errors.

src/ckb-daemon/usb_linux.c
```diff
@@ -67,7 +68,11 @@ static char kbsyspath[DEV_MAX][FILENAME_MAX];
 ///
 int os_usbsend(usbdevice* kb, const uchar* out_msg, int is_recv, const char* file, int line) {
     int res;
-    if ((kb->fwversion >= 0x120 || IS_V2_OVERRIDE(kb)) && !is_recv){
+    // If we need to read a response, lock the mutex so that os_usbrecv() waits for the main thread.
+    if(is_recv)
+        pthread_mutex_lock(&kb->interruptmutex);
```
light2yellow: Consider checking for errors.

src/ckb-daemon/usb_linux.c
light2yellow: Up to you, I am just reviewing the diff.

src/ckb-daemon/usb_linux.c
light2yellow: ... here, and in other pthread_* functions, no point in commenting under every call.

PR #107, opened 2018-02-24
Add Dark Core IDs

ZirconiumX:
These IDs aren't plumbed into usb_*.c, so they won't work until we get a volunteer with the device itself.


taylorcoffelt:
Just purchased one of these, how can I help?

ZirconiumX:
The fastest way to help would be to join us in IRC, by pressing the blue
Freenode button at the top of the readme.

PR #108, opened 2018-02-24
Expose physical keyboard layout for the GUI

ZirconiumX:
This attempts to expose the mined keyboard layout byte from the daemon for the GUI to use.


ZirconiumX:
Yeah, I should import my .vimrc onto my PC, but I'll reformat those lines.

ZirconiumX:
Since you're probably the one who will work on the GUI, that's your call, Tasos.

PR #110, opened 2018-02-25
added support for runit init system

paulemurray:
Tested on Void Linux


ZirconiumX:
This seems more like a Void-specific thing than a runit-specific thing.

paulemurray:
What makes is specific to void?

tatokis:
Sorry for the delay.
If you are still interested, please rebase against master with cmake.

PR #160, opened 2018-04-22
Strafe cleanup

tatokis:
This PR removes unnecessary "switch to SW mode" commands that were sent on every strafe sidelight update.
It also doesn't send sidelight updates on non-Strafe fullrange keyboards.

Some testing would be appreciated.

### DIFF
--- a/src/daemon/device_keyboard.c
+++ b/src/daemon/device_keyboard.c
@@ -30,13 +30,13 @@ int setactive_kb(usbdevice* kb, int active){
     pthread_mutex_unlock(imutex(kb));
 
     uchar msg[3][MSG_SIZE] = {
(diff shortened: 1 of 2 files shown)


SethBarberee:
I can test this in the coming days with my Strafe RGB

tatokis:
@SethBarberee sorry, it was tested by @ZirconiumX and it has a few issues.
I need to fix it first

ZirconiumX:
We ended up combining this commit with some refactoring to fix sidelights on the 1b13 Strafe, which is not fullrange and thus doesn't have the sidelights modified.

ZirconiumX:
Fixed in 86f0c086.

PR #194, opened 2018-06-23
Rework input/mac USB code

tatokis:
Now mac and linux input code is shared.
Input logic has been moved to keymap.c and functions have been split.

macOS USB code was also necessarily modified to handle interrupts as handled by the shared input code.


src/daemon/usb_mac.c
```diff
@@ -408,8 +374,9 @@ extern void keyretrigger(CFRunLoopTimerRef timer, void* info);
 
 void* os_inputmain(void* context){
     usbdevice* kb = context;
+
     int index = INDEX_OF(kb, keyboard);
-    // Monitor input transfers on all endpoints for non-RGB devices
+    // Monitor input transfers on all endpoints for non-legacy devices
```
tatokis: This should be "for legacy" and "For non-legacy" in the next line

PR #203, opened 2018-07-02
Rework macro tab

tatokis:
This PR changes the layout and the widgets used in the macro tab, fixing rendering issues on mac and other themes.

Before:
f7eca1fff41f

After:
49e875d27636

Also fixes the bottom of the `tip` under the program tab being cut off.


tatokis:
Need to change Macro Label to Macro Sequence, or Macro Actions and make it show keystrokes that don't register in a text input field, such as <esc>

PR #206, opened 2018-07-08
Created a Docker image that creates ckb_next debs for Ubuntu 18.04

Kedstar99:
Hello,
I am a young software developer looking to get in some contributions to open code. I have created a Dockerscript that builds ckb-next in deb form for Ubuntu 18.04. The packages recommended came from looking up ldd on ckb-next and ckb-next-daemon and using apt-file to find corresponding dependencies.

Ideally, this can be used in a pipeline or something like that to enable people to install ckb-next and hopefully this can be added to Ubuntu universe. I would love to have comments or changes if that is necessary.

Kind regards,
Krish

### DIFF
--- a/linux/debian/DOCKER/Dockerfile
+++ b/linux/debian/DOCKER/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:18.04
+LABEL maintainer="Krish <K.DeSouza@Outlook.com>"
+WORKDIR /ckb-next
+RUN apt-get update && apt-get install -y bash dpkg systemd git build-essential cmake libudev-dev qt5-default zlib1g-dev libappindicator-dev libpulse-dev libquazip5-dev
+COPY . /ckb-next
+ARG LC_ALL
+ENV LC_ALL C
+RUN mkdir -p /run/systemd/system
+RUN rm -rf ./build && \
+    cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -DSAFE_INSTALL=OFF -DSAFE_UNINSTALL=OFF && \
+    cmake --build build --target all -- -j 2 && \
+    cd build && \
+    cmake . && \
+    make package
+ENTRYPOINT ["sh",".linux/debian/DOCKER/entry_point.sh"]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,6 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules/sanitizer
 include(MacroEnsureOutOfSourceBuild)
 macro_ensure_out_of_source_build("ckb-next requires an out of source build. \
 Please create a separate build directory and run 'cmake /path/to/ckb-next [options]' there.")
-
 # Make sure we don't install in anaconda
 if("$ENV{PATH}" MATCHES "anaconda")
     string(REPLACE ":" ";" ENVLIST "$ENV{PATH}")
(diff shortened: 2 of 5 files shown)


Kedstar99:
Thanks for your input, I will try and make the changes to maybe fit a bit better with the project.

Kedstar99:
I have managed to create a deb file using just CPack and Cmake and Docker. I hope this is slightly more acceptable for your build system. It also automatically generates dependencies to avoid any hard-coding. My only question at present is if it's necessary at present to install cmake files for the animations.

Also, I haven't quite figured out how to restart ckb-next-daemon.

ZirconiumX:
So, we just read through your code.

First of all, this is great work, and we're really impressed with how simple this is.

We do have a few nitpicks, and we do want to test this, but once those are fixed we'd be happy to merge this.

The 64 bit/32 bit detection code is fairly fragile, but we can't think of a better solution for it at the moment.

Additionally, this is a merge, not a rebase, and thus looks cluttered with the changes we've made since you opened this PR. For the sake of avoiding duplicate commits, could you either rebase or squash your changes?

>Also, I haven't quite figured out how to restart ckb-next-daemon.

Assuming you meant "start" instead of "restart":
```
systemctl enable ckb-next-daemon.service
systemctl start ckb-next-daemon.service
```

Kedstar99:
I think I squashed the commits in my branch, I hope this is ok. 

>Assuming you meant "start" instead of "restart":

In regards to this, I meant that at present the deb doesn't install and run the ckb-next-daemon immediately on start. There is a way of doing that manually having a post-init script, but i am not entirely sure how to integrate it with c-pack just yet. That should be solved by the user simply restarting the system until i figure that out.

Also, the deb doesn't currently start at startup unless someone ticks and unticks the startup.
I am also haven't really tested the system outside my machine, but so far it seems to work fine.

ZirconiumX:
According to [this](https://stackoverflow.com/a/7918913/1967929), you could use `CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA` to bundle the postinst script with it.

The squashed commit looks much cleaner; thank you!

Kedstar99:
The postinst script is now implemented and documentation updated with some additional details. If you would like, I can send a copy of the deb I create for testing but ideally I want to show that others can recreate my steps in docker.

Please do let me know if there are any other things that I can do.

In regards to the j flag for building CKB earlier, does 2 jobs seem reasonable? I think 2 cores is roughly the default by this point. Ideally, I am hoping that people don't need to build ckb-next manually and we can instead just pass the produced deb instead.

Kedstar99:
Hello,
It's been a while, can I know if everything is ok to proceed or is there anything else I need to do?
Kind regards,
Krish

ZirconiumX:
Sorry, me and @tatokis have been on holiday recently, and we only got back today  (yes, our comments were while on holiday).

Since @tatokis' computer works (it crashed and we couldn't SSH in), we will try to test this soon. Rest assured, once we test this, I would be happy to merge this.

Kedstar99:
Hi guys,
Just a brief update over the changes. I updated the CMake file to contain correct updated versioning and I fixed the postinit script to actually restart ckb-next-daemon.
Krish

tatokis:
Thanks. I had a look and the deb gets generated but the .desktop file doesn't have the correct paths. That is an issue on our build system, but I haven't had a look yet. Sorry.

Been a bit busy trying to release 0.3.1 and juggling other projects at once.

linux/debian/DOCKER/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:18.04
+LABEL maintainer="Krish <K.DeSouza@Outlook.com>"
+WORKDIR /ckb-next
+RUN apt-get update && apt-get install -y bash dpkg systemd git build-essential cmake libudev-dev qt5-default zlib1g-dev libappindicator-dev libpulse-dev libquazip5-dev
+COPY . /ckb-next
+ARG LC_ALL
+ENV LC_ALL C
+RUN mkdir -p /run/systemd/system
+RUN rm -rf ./build && \
+    cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -DSAFE_INSTALL=OFF -DSAFE_UNINSTALL=OFF && \
```
Kedstar99: Is it because of this Install_prefix?

Kedstar99:
Is it because of the docker script setting prefix to /usr?

    cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -DSAFE_INSTALL=OFF -DSAFE_UNINSTALL=OFF && \

If that is the issue, I am not sure how to fix that bug.

tatokis:
Actually nevermind, it was a fault on my end. My Apologies.

Issue #212, opened 2018-07-30
link not created in /dev/input/by-id


m5x:
For evdev passthrough purposes it would be very handy if link to the keyboard device got created in /dev/input/by-id as usual for input devices. I am on Solus system where it does not get created, should I report this to Solus package maintainers or ckb-next does not do it yet?

tatokis:
Apparently I never linked this to the wiki, but https://github.com/ckb-next/ckb-next/issues/123#issuecomment-399744172

This should get you going. That said, I'm not sure what's responsible for creating these symlinks, but it's clearly not creating symlinks for uinput devices.
I'm not sure there's much we can do about it.

m5x:
I thought it gets created by udev rules file, usually called something like `60-persistent-input.rules`. There is already rules file in ckb-next repository, maybe the link creation could be added to it?

tatokis:
udev doesn't have enough data to create an identifiable by-path or a by-id symlink.

```
$ udevadm info /dev/input/event9
P: /devices/virtual/input/input469/event9
N: input/event9
E: BACKSPACE=guess
E: DEVNAME=/dev/input/event9
E: DEVPATH=/devices/virtual/input/input469/event9
E: ID_INPUT=1
E: ID_INPUT_KEY=1
E: ID_INPUT_KEYBOARD=1
E: ID_INPUT_MOUSE=1
E: ID_SERIAL=noserial
E: MAJOR=13
E: MINOR=73
E: SUBSYSTEM=input
E: USEC_INITIALIZED=349442600887
E: XKBLAYOUT=us,gr
E: XKBMODEL=pc105
E: XKBOPTIONS=grp:alt_shift_toggle,grp_led:scroll
E: XKBVARIANT=,

$ udevadm info --attribute-walk /dev/input/event9

Udevadm info starts with the device specified by the devpath and then
walks up the chain of parent devices. It prints for every device
found, all possible attributes in the udev rules key format.
A rule to match, can be composed by the attributes of the device
and the attributes from one single parent device.

  looking at device '/devices/virtual/input/input469/event9':
    KERNEL=="event9"
    SUBSYSTEM=="input"
    DRIVER==""

  looking at parent device '/devices/virtual/input/input469':
    KERNELS=="input469"
    SUBSYSTEMS=="input"
    DRIVERS==""
    ATTRS{name}=="ckb1: Corsair K70 RGB Gaming Keyboard"
    ATTRS{phys}==""
[...]

PR #217, opened 2018-08-04
Update udev rules to create symlinks for devices

urmamasllama:
This change addresses issue [212](https://github.com/ckb-next/ckb-next/issues/212)
This change creates symlinks in /dev/input/by-id to the event objects for each device. Each ckb device has one mouse and one keyboard event that have the same name and little to distinguish them from each other, so to handle this I appended the EV count to the end of the name. 7 is mouse and 120003 is keyboard. Since these event objects are technically not the actually device but more of a fake device to facilitate remapping I felt it was appropriate to leave `ckb` at the front of the name rather than the traditional `usb`.

This is the first time I've ever done a pull request so if I messed up in protocol somewhere I apologize.

### DIFF
--- a/linux/udev/99-ckb-daemon.rules
+++ b/linux/udev/99-ckb-daemon.rules
@@ -1,4 +1,4 @@
-# Create base ckb-next devices
+# Mark ckb devices as not a joystick
 ACTION=="add|change", SUBSYSTEM=="input", ATTRS{name}=="ckb[0-9]: [A-Za-z0-9]*", ENV{ID_INPUT_JOYSTICK}=""
 
 # If the device isn't an event input device with a recognised
--- a/linux/udev/99-ckb-daemon.rules
+++ b/linux/udev/99-ckb-daemon.rules
@@ -1,26 +1,2 @@
 # Mark ckb devices as not a joystick
-ACTION=="add|change", SUBSYSTEM=="input", ATTRS{name}=="ckb[0-9]: [A-Za-z0-9]*", ENV{ID_INPUT_JOYSTICK}=""
-
-# If the device isn't an event input device with a recognised
-# class, do nothing
-KERNEL!="event*", GOTO="input_ckb_ids_end"
-SUBSYSTEM!="input", GOTO="input_ckb_ids_end"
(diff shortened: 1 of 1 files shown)


tatokis:
>7 is mouse and 120003 is keyboard.

I really don't think that's a reliable way of detecting it.
Try using `$env{.INPUT_CLASS}`, which should either print `kbd` or `mouse`

tatokis:
Note to self, should also check what happens if the rule is installed on systems without systemd.

urmamasllama:
Unfortunately event class doesn't work. Just double checked in case I was missing something but if you use that instead only the mouse event gets a link. I've confirmed with four ckb devices so far that The virtual events use 7 and120003 even if that changes for anything using the ev value should work fine. We just wouldn't be able to use it as a catch to label instead.

urmamasllama:
so for reference here is examples of what these options would look like.
This is how the symlinks look with my current proposed changes
```ckb-Corsair_Gaming_SCIMITAR_PRO_RGB_Mouse-event-120003
ckb-Corsair_Gaming_SCIMITAR_PRO_RGB_Mouse-event-7
ckb-Corsair_K95W_Gaming_Keyboard-event-120003
ckb-Corsair_K95W_Gaming_Keyboard-event-7
```
Here is using `INPUT_CLASS`
```ckb-Corsair_Gaming_SCIMITAR_PRO_RGB_Mouse-event-mouse
ckb-Corsair_K95W_Gaming_Keyboard-event-mouse
```
and for completeness here is using both
```ckb-Corsair_Gaming_SCIMITAR_PRO_RGB_Mouse-event-120003-mouse
ckb-Corsair_Gaming_SCIMITAR_PRO_RGB_Mouse-event-7-mouse
ckb-Corsair_K95W_Gaming_Keyboard-event-120003-mouse
ckb-Corsair_K95W_Gaming_Keyboard-event-7-mouse
```

ZirconiumX:
This is unrelated, but what the heck is a K95W?

urmamasllama:
It's the original K95 18 macro buttons but no RGB it has white leds instead.

PR #221, opened 2018-08-08
Fixed issues presented by GCC 7 for devnode.c

Kedstar99:
Hello back again,
I kind of wanted to get into some more open coding. As such, I figured an easy place to start may be to address some of the warnings and errors presented during compilation. 

I hope this patch is ok. Please let me know if there is anything else I can do.

Kind regards,
Krish


Kedstar99:
For reference, these issues seem to be fixed:

[ 11%] Building C object src/daemon/CMakeFiles/ckb-next-daemon.dir/devnode.c.o
In file included from /ckb-next/src/animations/wave/main.c:1:0:
/ckb-next/src/libs/ckb-next/include/ckb-next/animation.h: In function 'ckb_getline':                                                                                                                                                                                           
/ckb-next/src/libs/ckb-next/include/ckb-next/animation.h:241:5: warning: ignoring return value of 'fgets', declared with attribute warn_unused_result [-Wunused-result]                                                                                                        
     fgets(line, sizeof(line), stdin);                                                                                                                                                                                                                                         
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                          
/ckb-next/src/daemon/devnode.c: In function '_updateconnected':                                                                                                                                                                                                                
/ckb-next/src/daemon/devnode.c:78:9: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                   
         chown(cpath, 0, gid);                                                                                                                                                                                                                                                 
         ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                                  
/ckb-next/src/daemon/devnode.c: In function '_mknotifynode':                                                                                                                                                                                                                   
/ckb-next/src/daemon/devnode.c:105:9: warning: ignoring return value of 'fchown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
         fchown(kb->outfifo[notify] - 1, 0, gid);                                                                                                                                                                                                                              
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                               
/ckb-next/src/daemon/devnode.c: In function '_mkdevpath':                                                                                                                                                                                                                      
/ckb-next/src/daemon/devnode.c:183:9: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                  
         chown(path, 0, gid);                                                                                                                                                                                                                                                  
         ^~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                                   
/ckb-next/src/daemon/devnode.c:197:17: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
                 chown(vpath, 0, gid);                                                                                                                                                                                                                                         
                 ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                          
/ckb-next/src/daemon/devnode.c:211:17: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
                 chown(vpath, 0, gid);                                                                                                                                                                                                                                         
                 ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                          
/ckb-next/src/daemon/devnode.c:230:13: warning: ignoring return value of 'fchown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                
             fchown(kb->infifo - 1, 0, gid);                                                                                                                                                                                                                                   
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                    
/ckb-next/src/daemon/devnode.c:278:17: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
                 chown(fpath, 0, gid);                                                                                                                                                                                                                                         
                 ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                          
/ckb-next/src/daemon/devnode.c: In function 'printnode':                                                                                                                                                                                                                       
/ckb-next/src/daemon/devnode.c:145:13: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
             chown(path, 0, gid);                                                                                                                                                                                                                                              
             ^~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                               
/ckb-next/src/daemon/devnode.c: In function 'rmdevpath':                                                                                                                                                                                                                       
/ckb-next/src/daemon/devnode.c:304:9: warning: ignoring return value of 'write', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                  
         write(fd, "\n", 1); // hack to prevent the FIFO thread from perma-blocking                                                                                                                                                                                            
         ^~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                                    
/ckb-next/src/daemon/devnode.c: In function 'mkfwnode':                                                                                                                                                                                                                        
/ckb-next/src/daemon/devnode.c:333:13: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
             chown(fwpath, 0, gid);                                                                                                                                                                                                                                            
             ^~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                             
/ckb-next/src/daemon/devnode.c:348:13: warning: ignoring return value of 'chown', declared with attribute warn_unused_result [-Wunused-result]                                                                                                                                 
             chown(ppath, 0, gid);                                                                                                                                                                                                                                             
             ^~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                                                              
/ckb-next/src/daemon/devnode.c: In function 'rmdevpath':                                                                                                                                                                                                                       
/ckb-next/src/daemon/devnode.c:121:54: warning: '__builtin___snprintf_chk' output may be truncated before the last format character [-Wformat-truncation=]                                                                                                                     
     snprintf(outpath, sizeof(outpath), "%s%d/notify%d", devpath, index, notify);                                                                                                                                                                                              
                                                      ^                                                                                                                                                                                                                        
In file included from /usr/include/stdio.h:862:0,                                                                                                                                                                                                                              
                 from /ckb-next/src/daemon/includes.h:12,                                                                                                                                                                                                                      
                 from /ckb-next/src/daemon/device.h:4,                                                                                                                                                                                                                         
                 from /ckb-next/src/daemon/devnode.c:1:                                                                                                                                                                                                                        
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:10: note: '__builtin___snprintf_chk' output between 24 and 35 bytes into a destination of size 24                                                                                                                               
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,                                                                                                                                                                                                         
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                                                         
        __bos (__s), __fmt, __va_arg_pack ());                                                                                                                                                                                                                                 
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/daemon/devnode.c
```diff
@@ -36,6 +36,33 @@ int rm_recursive(const char* path){
     return remove(path);
 }
 
+int check_chown(const char *pathname, uid_t owner, gid_t group) {
+    if (chown(pathname, owner, group)) {
```
ZirconiumX: As far as I can tell, this condition is wrong; `chown` returns 0 on success, but -1 on failure, so this path is taken when `chown` fails, and the failure path is taken when it succeeds.

Additionally, I would rather have this function worded like this:
```c
if (function() < 0) {
    ckb_warn("Oh noes!\n");
    return -1;
}
return 0;
```

src/daemon/devnode.c
```diff
@@ -36,6 +36,33 @@ int rm_recursive(const char* path){
...
+        return 0;
+    } else {
+        ckb_warn("Chown call failed %s: %s\n", pathname, strerror(errno));
+        return -1;
+    }
+}
+
+int check_fchown(int fd, uid_t owner, gid_t group) {
+    if (fchown(fd, owner, group)) {
```
ZirconiumX: See comments for `chown`, but also, why do we have both `chown` and `fchown`?

src/daemon/devnode.c
```diff
@@ -36,6 +36,33 @@ int rm_recursive(const char* path){
...
+        return 0;
+    } else {
+        ckb_warn("FChown call failed: %s\n", strerror(errno));
+        return -1;
+    }
+}
+
+int check_chmod(const char *pathname, mode_t mode) {
+    if(chmod(pathname, mode)) {
```
ZirconiumX: See comments for `chown`. I'm not sure a warning suffices in this particular case, though; I'm not familiar with this particular code, but I'd lean towards making this fatal for possible security implications.

src/daemon/devnode.c
```diff
@@ -73,9 +100,11 @@ void _updateconnected(){
     if(!written)
         fputc('\n', cfile);
     fclose(cfile);
-    chmod(cpath, S_GID_READ);
+    
+    (void)check_chmod(cpath, S_GID_READ);
```
ZirconiumX: I wouldn't call this any better than the original code. Casting it to `void` is not error handling.

src/daemon/devnode.c
```diff
@@ -92,7 +121,7 @@ int _mknotifynode(usbdevice* kb, int notify){
         return 0;
     // Create the notification node
     int index = INDEX_OF(kb, keyboard);
-    char outpath[strlen(devpath) + 10];
+    char outpath[strlen(devpath) + 8*sizeof(char) + 2*sizeof(int)];
```
ZirconiumX: I think you've misunderstood the code here; the original code was correct because `index` is always between `1` and `9`.

Your code is incorrect even under the assumption this invariant does not hold; the largest string a 32-bit integer can hold is `-2147483648` which is 11 characters, not 4, let alone a 64-bit int (which is rare), which has the longest string of `-9223372036854775808`, which is 20 characters long.

src/daemon/devnode.c
```diff
@@ -101,8 +130,9 @@ int _mknotifynode(usbdevice* kb, int notify){
         remove(outpath);
         return -1;
     }
-    if(gid >= 0)
-        fchown(kb->outfifo[notify] - 1, 0, gid);
+    if(gid >= 0) {
+        return check_fchown(kb->outfifo[notify] - 1, 0, gid);
```
ZirconiumX: Is there ever a case where `gid` isn't >= 0? If not, this should just be `return check_fchown(...)`.

src/daemon/devnode.c
```diff
@@ -117,7 +147,7 @@ int _rmnotifynode(usbdevice* kb, int notify){
     if(notify < 0 || notify >= OUTFIFO_MAX || !kb->outfifo[notify])
         return -1;
     int index = INDEX_OF(kb, keyboard);
-    char outpath[strlen(devpath) + 10];
+    char outpath[strlen(devpath) + 8*sizeof(char) + 2*sizeof(int)];
```
ZirconiumX: See comments for `outpath` in `_mknotifynode`.

src/daemon/devnode.c
```diff
@@ -140,9 +170,9 @@ static void printnode(const char* path, const char* str){
         fputs(str, file);
         fputc('\n', file);
         fclose(file);
-        chmod(path, S_GID_READ);
+        (void)check_chmod(path, S_GID_READ);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -140,9 +170,9 @@ static void printnode(const char* path, const char* str){
...
         if(gid >= 0)
-            chown(path, 0, gid);
+            (void)check_chown(path, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -180,7 +210,7 @@ static int _mkdevpath(usbdevice* kb){
         return -1;
     }
     if(gid >= 0)
-        chown(path, 0, gid);
+        (void)check_chown(path, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -192,9 +222,9 @@ static int _mkdevpath(usbdevice* kb){
         if(vfile){
             fprintf(vfile, "%s\n", CKB_NEXT_VERSION_STR);
             fclose(vfile);
-            chmod(vpath, S_GID_READ);
+            (void)check_chmod(vpath, S_GID_READ);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -192,9 +222,9 @@ static int _mkdevpath(usbdevice* kb){
...
             if(gid >= 0)
-                chown(vpath, 0, gid);
+                (void)check_chown(vpath, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -206,9 +236,9 @@ static int _mkdevpath(usbdevice* kb){
         if(pfile){
             fprintf(pfile, "%u\n", getpid());
             fclose(pfile);
-            chmod(ppath, S_READ);
-            if(gid >= 0)
-                chown(vpath, 0, gid);
+            (void)check_chmod(ppath, S_READ);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -206,9 +236,9 @@ static int _mkdevpath(usbdevice* kb){
...
+            if (gid >= 0)
+                check_chown(vpath, 0, gid);
```
ZirconiumX: You're not even bothering to cast to `void` here.

src/daemon/devnode.c
```diff
@@ -227,7 +257,7 @@ static int _mkdevpath(usbdevice* kb){
             return -1;
         }
         if(gid >= 0)
-            fchown(kb->infifo - 1, 0, gid);
+            (void)check_fchown(kb->infifo - 1, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -273,9 +303,9 @@ static int _mkdevpath(usbdevice* kb){
                 fputs(" hwload", ffile);
             fputc('\n', ffile);
             fclose(ffile);
-            chmod(fpath, S_GID_READ);
+            (void)check_chmod(fpath, S_GID_READ);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -273,9 +303,9 @@ static int _mkdevpath(usbdevice* kb){
...
             if(gid >= 0)
-                chown(fpath, 0, gid);
+                (void)check_chown(fpath, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -297,12 +327,12 @@ int rmdevpath(usbdevice* kb){
...
     if(kb->infifo != 0){
-        int fd = kb->infifo - 1;
+        FILE* fd = fdopen(kb->infifo - 1, "a");
 #ifdef OS_MAC
-        fcntl(fd, F_SETFL, O_RDWR | O_NONBLOCK); // hack to prevent the following hack from blocking if the GUI was running
+        fcntl(kb->infifo - 1, F_SETFL, O_RDWR | O_NONBLOCK); // hack to prevent the following hack from blocking if the GUI was running
 #endif
-        write(fd, "\n", 1); // hack to prevent the FIFO thread from perma-blocking
-        close(fd);
+        fputc('\n', fd); //hack to prevent the FIFO thread from perma-blocking
```
ZirconiumX: This is suspicious to me. The reason the `write` call was here is because it bypasses libc buffering; while it should get flushed with the `fclose`, I'm unsure about changing this code for change's sake.

src/daemon/devnode.c
```diff
@@ -297,12 +327,12 @@ int rmdevpath(usbdevice* kb){
     euid_guard_start;
     int index = INDEX_OF(kb, keyboard);
     if(kb->infifo != 0){
-        int fd = kb->infifo - 1;
+        FILE* fd = fdopen(kb->infifo - 1, "a");
```
ZirconiumX: This isn't actually an improvement on the original code; you go to the trouble of making a `FILE*` here, and then immediately use the old `int` fd in the `fcntl` call.

src/daemon/devnode.c
```diff
@@ -328,9 +358,9 @@ int mkfwnode(usbdevice* kb){
         fprintf(fwfile, "%04x", kb->fwversion);
         fputc('\n', fwfile);
         fclose(fwfile);
-        chmod(fwpath, S_GID_READ);
+        (void) check_chmod(fwpath, S_GID_READ);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

I shouldn't have to copy and paste this again and again.

src/daemon/devnode.c
```diff
@@ -328,9 +358,9 @@ int mkfwnode(usbdevice* kb){
...
         if(gid >= 0)
-            chown(fwpath, 0, gid);
+           (void)check_chown(fwpath, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -343,9 +373,9 @@ int mkfwnode(usbdevice* kb){
         fprintf(pfile, "%d ms", kb->pollrate);
         fputc('\n', pfile);
         fclose(pfile);
-        chmod(ppath, S_GID_READ);
+        (void)check_chmod(ppath, S_GID_READ);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

src/daemon/devnode.c
```diff
@@ -343,9 +373,9 @@ int mkfwnode(usbdevice* kb){
...
         if(gid >= 0)
-            chown(ppath, 0, gid);
+            (void)check_chown(ppath, 0, gid);
```
ZirconiumX: See comments for `check_chmod` in `_updateconnected`.

Kedstar99:
Thanks for your input, Im a little rusty in my C . I will try and fix this and get back to you for a patch.

src/daemon/devnode.c
Kedstar99: I kind of wanted to use the FILE* and fputc approach because the code only requires one character to be written and doesn't require all the error handling that was necessary for the write system call.

To add to this, is it a bug in this case to open and then do fopen? Would that leave two dangling file descriptors?

Kedstar99:
Hello, 
So I changed some of the code to fit your comments.
--Now if a function fails chmod, chown or fchown I just exit. That seems reasonable because if you can't perform those actions it seems like it would require user intervention. If otherwise please suggest.
--For the snprintf issue, I have code above that guarantees the int is between 0-9 and cast it to a char. This solves the warning issues
--For the write issue, I just add error handling for that specific problem.
--I have avoided the casting to void by changing the return types for those functions to void.

Kind regards,
Krish

Kedstar99:
By the way, I just checked some static analysis on some of the shell scripts I have been writing.

It turns out there is a mistake in the shell script in ./linux/debian/postinst

The shebang should have an exclamation a la

#!/bin/bash

It's minor, just worth correcting in a small commit.

Kedstar99:
Same with ./linux/debian/DOCKER/entry_point.sh

src/daemon/devnode.c
```diff
@@ -301,7 +324,9 @@ int rmdevpath(usbdevice* kb){
 #ifdef OS_MAC
         fcntl(fd, F_SETFL, O_RDWR | O_NONBLOCK); // hack to prevent the following hack from blocking if the GUI was running
 #endif
-        write(fd, "\n", 1); // hack to prevent the FIFO thread from perma-blocking
+        int write_ret = write(fd, "\n", 1); // hack to prevent the FIFO thread from perma-blocking        fclose(fd);
+        if (write_ret < 0)
+            ckb_warn("Unable to write to filedescriptor %d: %s\n", fd, strerror(errno));
```
ZirconiumX: Couldn't this be simplified to `if (write(fd, "\n", 1) < 0)`?

Kedstar99:
Fixed the change.

PR #234, opened 2018-08-19
Improved compliance for ISO C for the codebase

Kedstar99:
Hello again,
I have again done some general minor fixes across the code base. Mainly in regards to some of the errors shown in the pedantic flags.

Some of these errors are likely to cause undefined or weird behaviour. Such as on clang the use of designated initializers as that is a GCC only extension.

Mostly just fixed issues with respect to sscanf, but at least now the code compiles under pedantic and seems to function.

I hope these changes are useful.

Kind regards,
Krish

### DIFF
--- a/src/daemon/profile.c
+++ b/src/daemon/profile.c
@@ -3,6 +3,7 @@
 #include "input.h"
 #include "led.h"
 #include "profile.h"
+#include "stdint.h"
 
 // Percent-enconding conversions
 void urldecode2(char* dst, const char* src){
@@ -62,9 +63,9 @@ void urlencode2(char* dst, const char* src){
 
 // GUID conversions
 int setid(usbid* id, const char* guid){
-    int32_t data1;
-    int16_t data2, data3, data4a;
-    char data4b[6];
+    uint data1;
+    ushort data2, data3, data4a;
+    uchar data4b[6];
     if(sscanf(guid, "{%08X-%04hX-%04hX-%04hX-%02hhX%02hhX%02hhX%02hhX%02hhX%02hhX}",
               &data1, &data2, &data3, &data4a, data4b, data4b + 1, data4b + 2, data4b + 3, data4b + 4, data4b + 5) != 10)
         return 0;
@@ -77,9 +78,9 @@ int setid(usbid* id, const char* guid){
 }
 
 char* getid(usbid* id){
-    int32_t data1;
-    int16_t data2, data3, data4a;
-    char data4b[6];
+    uint data1;
+    ushort data2, data3, data4a;
+    uchar data4b[6];
     memcpy(&data1, id->guid + 0x0, 4);
     memcpy(&data2, id->guid + 0x4, 2);
     memcpy(&data3, id->guid + 0x6, 2);
@@ -171,7 +172,7 @@ void cmd_id(usbdevice* kb, usbmode* mode, int dummy1, int dummy2, const char* id
     (void)dummy2;
 
     // ID is either a GUID or an 8-digit hex number
-    int newmodified;
+    uint newmodified;
     if(!setid(&mode->id, id) && sscanf(id, "%08x", &newmodified) == 1)
         memcpy(mode->id.modified, &newmodified, sizeof(newmodified));
 }
@@ -182,7 +183,7 @@ void cmd_profileid(usbdevice* kb, usbmode* mode, int dummy1, int dummy2, const c
     (void)dummy2;
 
     usbprofile* profile = kb->profile;
-    int newmodified;
+    uint newmodified;
     if(!setid(&profile->id, id) && sscanf(id, "%08x", &newmodified) == 1)
         memcpy(profile->id.modified, &newmodified, sizeof(newmodified));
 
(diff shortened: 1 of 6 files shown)


DarthVaderx99:
thanks

src/daemon/device.c
```diff
@@ -4,18 +4,19 @@
...
 pthread_mutex_t devlistmutex = PTHREAD_MUTEX_INITIALIZER;
-pthread_mutex_t devmutex[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_MUTEX_INITIALIZER };      ///< Mutex for handling the usbdevice structure
-pthread_mutex_t inputmutex[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_MUTEX_INITIALIZER };    ///< Mutex for dealing with usb input frames
-pthread_mutex_t macromutex[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_MUTEX_INITIALIZER };    ///< Protecting macros against lightning: Both use usb_send
-pthread_mutex_t macromutex2[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_MUTEX_INITIALIZER };   ///< Protecting the single link list of threads and the macrovar
-pthread_cond_t macrovar[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_COND_INITIALIZER };        ///< This variable is used to stop and wakeup all macro threads which have to wait.
-pthread_mutex_t interruptmutex[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_MUTEX_INITIALIZER };///< Used for interrupt transfers
-pthread_cond_t interruptcond[DEV_MAX] = { [0 ... DEV_MAX-1] = PTHREAD_COND_INITIALIZER };   ///< Same as above
+pthread_mutex_t devmutex[DEV_MAX] = MUTEX_INIT;      ///< Mutex for handling the usbdevice structure
```
Kedstar99: This stuff, i am curious about the behaviour of clang vs gcc. I am trying to take a look if the extension is secretly supported on Clang anyway, but there doesn't appear to be in any documentation. In which case, is this code expressed or does this even work when compiled with clang?

tatokis:
It compiles in Mac clang for sure. We don't use gcc for Mac builds. No warnings emitted with the current flags.

ZirconiumX:
Clang should accept it with `-std=gnu11`.

Kedstar99:
Is that flag set explicitly in cmake?

ZirconiumX:
https://cmake.org/cmake/help/v3.1/prop_tgt/C_STANDARD.html

https://github.com/ckb-next/ckb-next/blob/master/src/daemon/CMakeLists.txt#L162

Not explicitly, but it should work.

Kedstar99:
Do you prefer the approach I followed instead? It's not the cleanest relative to solutions in other languages or this approach, but it prevents the warning with the pedantic flag which I think may be useful for catching other issues.

ZirconiumX:
I'm not entirely happy with it, but it works, which is good enough.

tatokis:
I personally prefer the existing initialiser. The MUTEX_INIT one seems more unreadable, and will also be more difficult to maintain in the future if we decide to raise DEV_MAX.

src/gui/gradientbutton.cpp
```diff
@@ -45,7 +45,7 @@ void GradientButton::fromString(const QString& string){
     uchar a, r, g, b;
     while(1){
         int scanned = 0;
-        char newpos;
+        signed char newpos;
```
tatokis: I'm not sure if this is needed, as we pass `-fsigned-char`

src/daemon/profile.c
```diff
@@ -62,9 +63,9 @@ void urlencode2(char* dst, const char* src){
 
 // GUID conversions
 int setid(usbid* id, const char* guid){
-    int32_t data1;
-    int16_t data2, data3, data4a;
-    char data4b[6];
+    uint32_t data1;
+    uint16_t data2, data3, data4a;
+    unsigned char data4b[6];
```
tatokis: For consistency, please use the typedefs when possible.
https://github.com/ckb-next/ckb-next/blob/master/src/daemon/includes.h#L24-L26

src/gui/gradientbutton.cpp
tatokis: I take that back. We only pass it to the daemon.

Kedstar99:
Sorry it's been a while since I looked at this code. I have kind of been working on other projects.

Is there anything I need to do for this request or do i need to manually revert changes? If so i'll try and look at it sometime this week.

tatokis:
I could, but since it is your commit and you're still here, I would rather not mess with it.
You can revert the mutex/cond change manually, amend the commit, and force push this branch.

Kedstar99:
Done

Kedstar99:
Sorry will fix the commit message and device.c and new line change.

src/gui/keymap.cpp
```diff
@@ -108,10 +108,10 @@ static const KeyPatch patchSE[] = {
     {"angle", "<", "bslash_iso"}, {"minus", "-", "slash"},
 };
 
-static const KeyPatch patchJP[] = {
+/*static const KeyPatch patchJP[] = {
     /*{"lbrace", "[", "rbrace"},
```
tatokis: Apologies, one more thing I noticed just now. Please remove this `/*` that's inside the comment.

Kedstar99:
Anything else I need to do with this?

PR #238, opened 2018-08-24
Invaders

tatokis:
This PR adds and reworks the invaders animation by @sh4d0w28 that was queued up to be merged in the original ckb, but was only noticed recently by @ZirconiumX.
https://github.com/ccMSC/ckb/pull/553

A simple debugging function was also added to the GUI, however if needed it can be split to a separate PR.

It needs to be merged after the license issue is cleared up, as it is GPLv3.
ckb-next should be GPLv2 or later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ cmake_dependent_option(WITH_RANDOM   "Build with 'Random' animation."   ON "WITH
 cmake_dependent_option(WITH_PINWHEEL "Build with 'Pinwheel' animation." ON "WITH_ANIMATIONS;WITH_GUI" OFF)
 cmake_dependent_option(WITH_RIPPLE   "Build with 'Ripple' animation."   ON "WITH_ANIMATIONS;WITH_GUI" OFF)
 cmake_dependent_option(WITH_WAVE     "Build with 'Wave' animation."     ON "WITH_ANIMATIONS;WITH_GUI" OFF)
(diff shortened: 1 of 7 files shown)


sh4d0w28:
wow. I just saw your commit. Thanks for help! Also, what about license version? i think we still can downgrade my repository version to GPLv2 ?

ZirconiumX:
ckb-next is GPLv2 or later, we just need you to agree to relicensing your
code to v2, and we can add it to the codebase.

You're lucky I stumbled upon it, or it might have gotten lost to the mists
of time altogether. It's surprising it still works after 5 new releases.

tatokis:
The issue is that the animation is GPLv3.
If ckb-next is GPLv2 only, then we can not merge the code.
If ckb-next is GPLv2 or later, then we can merge the code as it will be compatible under the "or later" part.

We have been going with the assumption that ckb-next is GPLv2 only, but looking at MSC's repo, I don't see evidence to support that. We need to sort this out ASAP.
If we do decide to go by "GPLv2 or later", then we need to explicitly mention that somewhere.

@sh4d0w28 if you're willing to relicense to GPLv2, then this will make the process easier.
If so, fork the repo, checkout the invaders branch, and make a commit to your fork with *the same commit credentials* as your previous commits.

sh4d0w28:
I did in commit
https://github.com/ckb-next/ckb-next/pull/248

sh4d0w28:
meanwhile, if you need, you have my approval on downgrading license version to v2.

tatokis:
Thank you

PR #239, opened 2018-08-24
K66

tatokis:
This PR adds support for the K66.
It has no backlight but still uses the CUE protocol.

### DIFF
--- a/src/gui/keymap.h
+++ b/src/gui/keymap.h
@@ -48,9 +48,10 @@ struct Key {
 // Key layout/device info class
 class KeyMap {
 public:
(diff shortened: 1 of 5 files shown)


src/gui/keymap.h
```diff
@@ -48,9 +48,10 @@ struct Key {
 // Key layout/device info class
 class KeyMap {
 public:
+    // Do not reorder these.
+    // Doing so will break the profile import/export
```
ZirconiumX: Note for any future revamp of the profile code to not make the key map dependent on a hardcoded value.

PR #246, opened 2018-08-29
Clean up FindIconv.cmake test

pearsonk:
The iconv.h header's iconv function does not take (const char **)
as its second argument, rather a (char **). Fixing the test
declaration allows paranoid compilers to continue.

Ref: http://pubs.opengroup.org/onlinepubs/7908799/xsh/iconv.h.html

Signed-off-by: Kevin Pearson <kevin.pearson@ortmanconsulting.com>

### DIFF
--- a/cmake/modules/FindIconv.cmake
+++ b/cmake/modules/FindIconv.cmake
@@ -78,7 +78,7 @@ IF(ICONV_FOUND)
   #include <iconv.h>
   int main(){
     iconv_t conv = 0;
-    const char* in = 0;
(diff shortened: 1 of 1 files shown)


tatokis:
@fleischie Fair point. Should we close the PR, then?

fleischie:
@tatokis If the only intention was to potentionally fix paranoid compilers as the description indicates: Yes.

If there is a deep philosophical reason: Maybe.

tatokis:
Fair enough then

PR #213, opened 2018-08-01
Rename UDEV rules file

dguglielmi:
Use same file naming as other files (systemd unit, openrc init, etc..)

PR #269, opened 2018-10-08
Rename udev rule without changes

Kedstar99:
Renamed the udev rule as per https://github.com/ckb-next/ckb-next/issues/213


src/daemon/CMakeLists.txt
```diff
@@ -454,8 +454,11 @@ elseif ("${CKB_NEXT_INIT_SYSTEM}" STREQUAL "unknown")
 endif ()
 
 if (LINUX)
+  if (SAFE_INSTALL)
+    file(RENAME "/etc/udev/rules.d/99-ckb-daemon.rules" "/etc/udev/rules.d/99-ckb-next-daemon.rules")
+  endif
```
tatokis:
```
CMake Error at src/daemon/CMakeLists.txt:460:
  Parse error.  Expected "(", got newline with text "

  ".


-- Configuring incomplete, errors occurred!
See also "build/CMakeFiles/CMakeOutput.log".
An error occurred,
press enter to exit.
```
This not being tested aside, it's supposed to be endif()

Kedstar99:
Fixed, sorry for not testing the change.

tatokis:
```
CMake Error at src/daemon/CMakeLists.txt:458 (file):
  file RENAME failed to rename

    /etc/udev/rules.d/99-ckb-daemon.rules

  to

    /etc/udev/rules.d/99-ckb-next-daemon.rules

  because: Permission denied
```
because it runs during configuration time without the appropriate permissions.

Kedstar99:
What would you suggest to be the solution to this issue?

As far as I am aware, I was told this patch was kind of wanted, Zirconium just suggested the file should be removed before hand.

Isn't this meant to be run with sudo priviledges when installing the file?

Kedstar99:
Tested and it now works with quick_install.sh

src/daemon/CMakeLists.txt
```diff
@@ -454,8 +454,11 @@ elseif ("${CKB_NEXT_INIT_SYSTEM}" STREQUAL "unknown")
 endif ()
 
 if (LINUX)
+  if (SAFE_INSTALL)
+    install(CODE "file(REMOVE \"/etc/udev/rules.d/99-ckb-daemon.rules\")")
```
light2yellow: At first I thought it's a clever way to migrate a user after breaking backwards compatibility.

However, depending on the resolution of #311 `/etc` is supposed to contain user-made udev rules, so this line cannot persist, since it would erase them upon each potential reinstallation.

tatokis: We could hash the file in its current state, and if it matches the hash, then cmake removes it and migrates it.

light2yellow: Good idea, `file(<MD5|SHA1|SHA224|SHA256|SHA384|SHA512> <filename> <variable>)` can do it.

Kedstar99: I can't edit this request right now as I'm on holiday presently with limited internet access. Do you need me to do anything as I'm currently not back till mid January.

tatokis: @Kedstar99 don't worry, we can take care of it. Apologies for holding this for so long

PR #273, opened 2018-10-10
Don't hardcode /usr/lib

ryanfitzsimon:
Use CMAKE_INSTALL_LIBDIR (provided by GNUInstallDirs module)
instead of hardcoding /usr/lib.

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -419,7 +419,7 @@ if ("${CKB_NEXT_INIT_SYSTEM}" STREQUAL "launchd")
 elseif ("${CKB_NEXT_INIT_SYSTEM}" STREQUAL "systemd")
     install(
         FILES "${CMAKE_CURRENT_BINARY_DIR}/service/ckb-next-daemon.service"
-        DESTINATION "/usr/lib/systemd/system"
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/systemd/system"
         PERMISSIONS
         OWNER_READ OWNER_WRITE
         GROUP_READ
--- a/src/libs/ckb-next/CMakeLists.txt
+++ b/src/libs/ckb-next/CMakeLists.txt
@@ -75,12 +75,12 @@ if(NOT MACOS)
         NAMESPACE
           ${CMAKE_PROJECT_NAME}::
         DESTINATION
-          "/usr/lib/cmake/${CMAKE_PROJECT_NAME}/${PROJECT_NAME}")
+          "${CMAKE_INSTALL_LIBDIR}/cmake/${CMAKE_PROJECT_NAME}/${PROJECT_NAME}")
 
     install(
         FILES
           "cmake/${PROJECT_NAME}Config.cmake"
           "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}/${PROJECT_NAME}ConfigVersion.cmake"
         DESTINATION
-          "/usr/lib/cmake/${CMAKE_PROJECT_NAME}/${PROJECT_NAME}")
+          "${CMAKE_INSTALL_LIBDIR}/cmake/${CMAKE_PROJECT_NAME}/${PROJECT_NAME}")
 endif()


light2yellow:
Consider adding this to the next release changelog, since it may require an [action](https://aur.archlinux.org/cgit/aur.git/commit/?h=ckb-next-git&id=eb78acddcadd8d588719d2b25e2e515b99126e6f) from packagers.

ryanfitzsimon:
> Consider adding this to the next release changelog

@light2yellow I agree, this should be noted in the changelog.

Did it actually break the ckb-next-git AUR package?
As far as I can tell, `CMAKE_INSTALL_LIBDIR` would be set to `/usr/lib64` which is symlinked to `/usr/lib` on Arch anyway. That is ugly though and `/usr/lib` is definitely the right place under Arch.

> in Fedora 29 x86_64 systemd unit files are in /usr/lib/systemd/system

@maage that's a valid concern. It looks like Fedora has a [packaging exception for systemd](https://pagure.io/packaging-committee/issue/158) which means following the normal convention won't work.

Am I correct in thinking that Fedora expects CMake find modules to be installed under `/usr/lib64/cmake`, so `CMAKE_INSTALL_DIR` still works for them?

> This looks like decent way to handle systemd unit files:
> https://github.com/ximion/limba/blob/master/data/cmake/systemdservice.cmake

This approach won't work for everyone because it returns an absolute path, ignoring the prefix.
On my system (NixOS) I get this: 
```bash
$ pkg-config --variable=systemdsystemunitdir systemd
/nix/store/pb1vvbzzm0q9axbcbv9989mbz10f881z-systemd-239/lib/systemd/system
```
That's a readonly directory, and definitely not the right place to try and install to.

 After a quick look at some Fedora spec files, it seems like it's quite common to just use `install` and `%{_unitdir}`, even for CMake projects.

Maybe the best thing to do would be to add a `CMAKE_INSTALL_SYSTEMD_SERVICEDIR` variable and use that like [rdma-core does](https://github.com/linux-rdma/rdma-core/blob/master/redhat/rdma-core.spec).

This could default to `CMAKE_INSTALL_LIBDIR` (which I believe works for most cases) and be overridden by packagers as needed.

light2yellow:
This patch by itself is correct. It fixes the places which I have overlooked while writing the initial listfiles, and implements it the way it's supposed to be. I but forgot to replace the full path with a variable (since I wrote this on Arch Linux machine and followed its packaging guidelines while discovering how packaging works, and testing it).

Current CMake implementation of ckb-next provides a set of variables:
* `GNUInstallDirs` module defines multiple CMake variables, described [here](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html)
* `CMAKE_INSTALL_PREFIX` CMake variables, described [here](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html)
* `DESTDIR` environment variable, described [here](https://cmake.org/cmake/help/latest/envvar/DESTDIR.html)
* `INSTALL_DIR_ANIMATIONS` CMake variable, described [here](https://github.com/ckb-next/ckb-next/wiki/CMake-Options)

A set of defaults has to be chosen, and it will be almost certainly wrong for every other distribution (not even mentioning ones with obscure file hierarchies like NixOS). For example, let's look at the output of `sudo cmake --build build --target install` on Arch Linux, configured with all the default options except `CMAKE_BUILD_TYPE=Release`:

```
-- Install configuration: "Release"
-- Installing: /usr/bin/ckb-next-dev-detect
-- Installing: /usr/bin/ckb-next-daemon
-- Installing: /usr/lib64/systemd/system/ckb-next-daemon.service
-- Installing: /etc/udev/rules.d/99-ckb-daemon.rules
-- Installing: /usr/lib64/cmake/ckb-next/Animation/AnimationTargets.cmake
-- Installing: /usr/lib64/cmake/ckb-next/Animation/AnimationConfig.cmake
-- Installing: /usr/lib64/cmake/ckb-next/Animation/AnimationConfigVersion.cmake
-- Installing: /usr/libexec/ckb-next-animations/gradient
-- Installing: /usr/libexec/ckb-next-animations/heat
-- Installing: /usr/libexec/ckb-next-animations/pinwheel
-- Installing: /usr/libexec/ckb-next-animations/rain
-- Installing: /usr/libexec/ckb-next-animations/random
-- Installing: /usr/libexec/ckb-next-animations/ripple
-- Installing: /usr/libexec/ckb-next-animations/wave
-- Installing: /usr/libexec/ckb-next-animations/invaders
-- Installing: /usr/libexec/ckb-next-animations/mviz
-- Installing: /usr/bin/ckb-next
-- Installing: /usr/share/applications/ckb-next.desktop
-- Installing: /usr/share/icons/hicolor/512x512/apps/ckb-next.png
```

* udev rules under `/etc/udev` -- OK for Arch Linux, but who knows to what `%{_udevrulesdir}` expands in SUSE and Red Hat distros
* libexec-like (according to [GNU Coding Standards](https://www.gnu.org/prep/standards/)) files under `libexec` -- violation of Arch Linux and openSUSE guidelines, but not a violation of Gentoo and Fedora guidelines
* RHEL package contains 256x256 icons instead of 512x512, no idea why

So, as a user, it's always better to aim for a distro-packaged version, because it's impossible to satisfy everyone's needs by upstream. Otherwise you are on your own and supposed to know better.

---

> Did it actually break the ckb-next-git AUR package?

See [the discussion](https://aur.archlinux.org/packages/ckb-next-git/). Looks like Manjaro is somewhat different (I was too lazy to figure it out, but I probably should), but since the change doesn't break Arch Linux installs, I decided to use `lib`. On Arch Linux all
`/usr/lib64/...` are resolved to `/usr/lib/...` either by filesystem or by pacman (I don't know much about symlinks).

---

> at least in Fedora 29 x86_64 systemd unit files are in /usr/lib/systemd/system

According to [Fedora systemd-specific packaging guidelines](https://fedoraproject.org/wiki/Packaging:Systemd#Filesystem_locations), it's `/lib/systemd/system`. However, e.g. Arch Linux has `/lib` linked to `/usr/lib`, so it might be same on Fedora. Might, because e.g. openSUSE has separate lib, lib32 and lib64, and its guidelines have yet another thing to say about [it](https://en.opensuse.org/openSUSE:Packaging_Conventions_RPM_Macros#.25_libdir) and [systemd-specific guidelines](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines) are present as well (it can be `/usr/lib/systemd/system`, but also something else, I assume depending on the package arch).

---

Considering the above, I don't think that a new variable is justified here, yet. If you can demonstrate a case, where merely setting `CMAKE_INSTALL_LIBDIR` and/or other currently available variables violates your distribution's guidelines or even worse, breaks it, and there is no other way other than fully specifying the systemd services installation path, then -- fair enough, a separate variable is required.

I don't think `CMAKE_INSTALL_LIBDIR` should be flipped to anything be default as well. You can read the source code of `GNUInstallDirs` module [here](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/GNUInstallDirs.cmake), it explains when and why a particular value is set.

tatokis:
I don't personally want to deal with this. This change appears to be breaking quickinstall for users, so I will revert it for now until someone can come up with a PR that does not have this issue.

ryanfitzsimon:
> I don't personally want to deal with this

@tatokis I'm happy to spend some time finding a solution that works for everyone.

> This change appears to be breaking quickinstall for users

Could you please elaborate on this so I can test it? Do you just mean the unit file isn't installed to the correct place?

PR #311, opened 2018-12-09
src/daemon: fix installation location of udev rules.

maxice8:
Install to /usr/lib/udev/rules.d since that is the correct location
for package-provided configuration.

/etc/udev/rules.d is reserved for local administrators that takes
precedence over /usr/lib/udev/rules.d when udev is searching for rules

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -456,7 +456,7 @@ endif ()
 if (LINUX)
   install(
     FILES "${CMAKE_SOURCE_DIR}/linux/udev/99-ckb-daemon.rules"
-    DESTINATION "/etc/udev/rules.d"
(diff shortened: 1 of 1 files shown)


tatokis:
I was working on merging this with #269, and just as I finished I realised something.
Ubuntu 16.04 doesn't read rules from /usr/lib/udev/rules.d/.

I will push the code to a branch, but I won't merge it to master until this gets resolved.
Would /lib/udev/rules.d/ be appropriate, instead?

tatokis:
> That must be some very very old udev to not read from /usr/lib, i'm suprised.

It's the one that comes with systemd 229.
That said, I still try to support Ubuntu 14.04, and that is pre-systemd.

> I see no problem with it.

Alright, I will update the location and merge to master. Thank you and apologies for taking this long.

Issue #329, opened 2019-01-02
Use pthread_setname_np


tatokis:
It would be a good idea to name the threads where possible so that it's easier to deal with when debugging.

Might also be a good idea to dynamically name the threads.
Something along the lines of "ckb1 input thread" will be okay.

PR #343, opened 2019-01-08
Named Threads

fleischie:
This PR addresses #329 

Use `pthread_setname_np` to supply names to threads to help differentiating them when debugging. This unfortunately needs to be separately done on BSD and Linux systems because the BSD version of this function only names the current thread and thus needs to be called from within the appropriate thread. The Linux version needs the thread to be named as its first argument and thus needs to be calls outside of the thread.

This PR makes the necessary distinctions between the mac and non-mac code and uses the above mentioned routine to name the threads.

Tested named usb- and input-threads on Arch Linux x86 kernel v4.20, and macOS Mojave. (I didn't test the macro threads, though.)

### DIFF
--- a/src/daemon/includes.h
+++ b/src/daemon/includes.h
@@ -66,4 +66,6 @@ void timespec_add(struct timespec* timespec, long nanoseconds);
 // Common structs
 #include "structures.h"
 
+#define THREAD_NAME_MAX 16
+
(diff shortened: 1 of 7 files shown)


tatokis:
I'll squash merge these and also add a warning in case we ever decide to support more than 9 devices.

PR #344, opened 2019-01-08
Renamed .rules file as well

luigino:


ZirconiumX:
This won't work, because it doesn't do anything about the aleady existing `99-ckb-daemon.rules`. Additionally, we already have #269, which this is a duplicate of.

PR #367, opened 2019-02-04
Conway's Game of Life Keyboard animation.

pa-sword:
Edits welcome.


pa-sword:
I'd prefer to be consulted if this animation needs to be relicensed, but you have my permission to relicense under GPLv2+.

tatokis:
Knowing almost nothing about this, I can sort of get this to work on the numpad (due to its layout) and understand what it's doing.

An issue I have is that some keys like <kbd>G</kbd> are flickering constantly instead of being on.

I'm okay with merging this, but I'd like for the flickering issue to be fixed first.

PR #369, opened 2019-02-07
Debianizing ckb-next

stewbond:
I'm packaging ckb-next for inclusion in Debian main repositories (and Debian-based distros like Ubuntu).  This pull request contains all of the patches I made to get this accepted into Debian.

These commits are all related in some way to Debian's lintian tags.

I've added man pages for each binary[1] and referenced it in the service file[2] 

I've added a CMake option for using FreeDesktop.org-standardized install paths (The ones used currently are only appropriate for RedHat and Arch-based distros).  This is important for systemd unit files[3], udev rules[4], and libexec[5].  Builders can now `-DDEBIANIZE_INSTALL_PATHS=ON` to install these files to debian-friendly paths.  It is disabled by default so your users will not experience any changes unless they set this.

I've removed a deprecated and unnecessary tag from desktop entry files (no functional changes)[6].

I fixed some spelling errors in binaries [7].

[1] https://lintian.debian.org/tags/binary-without-manpage.html
[2] https://lintian.debian.org/tags/systemd-service-file-missing-documentation-key.html
[3] https://lintian.debian.org/tags/systemd-service-file-outside-lib.html
[4] https://wiki.debian.org/udev
[5] https://lists.debian.org/debian-devel/2005/05/msg00412.html
[6] https://lintian.debian.org/tags/desktop-entry-contains-encoding-key.html
[7] https://lintian.debian.org/tags/spelling-error-in-binary.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,13 @@ if (MACOS)
     set(INSTALL_DIR_ANIMATIONS "ckb-next.app/Contents/Resources/${ANIMATIONS_DIR_NAME}"
         CACHE STRING "Where to install animations.")
 elseif (LINUX)
-    set(INSTALL_DIR_ANIMATIONS "${CMAKE_INSTALL_LIBEXECDIR}/${ANIMATIONS_DIR_NAME}"
+    option(DEBIANIZE_INSTALL_PATHS "Install files according to Debian and FreeDesktop.org policy" OFF)
+    if (${DEBIANIZE_INSTALL_PATHS})
+        set(_ANIMATION_PREFIX ${CMAKE_INSTALL_LIBDIR})
+    else()
+        set(_ANIMATION_PREFIX ${CMAKE_INSTALL_LIBEXECDIR})
+    endif()
+    set(INSTALL_DIR_ANIMATIONS "${_ANIMATION_PREFIX}/${ANIMATIONS_DIR_NAME}"
         CACHE STRING "Where to install animations.")
 endif ()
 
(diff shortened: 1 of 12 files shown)


ZirconiumX:
Thank you for your efforts, but you should have asked us before doing this.
 
Corsair have a tendency for unannounced breaking changes in their protocols, with no help provided to us for deciphering the resulting new protocols. Examples of this would be:
- the recent rollout of firmware version 3 changing the USB endpoint configuration.
- the K95 Platinum and MK.2 series introduced a new hardware profile format which we still have not deciphered, and so don't support due to legal issues of distributing an opaque binary blob.
- the Dark Core and K63 Wireless use a different firmware format requiring special update procedures.
- the Harpoon Wireless uses an entirely new protocol.
 
Corsair provide no way to downgrade firmware on Windows, which means a CUE firmware update could break compatibility with an older version of CKB.

In light of this, development of ckb-next usually moves much quicker than these distributions (especially Ubuntu), and we don't want to be receiving bug reports about older versions of our code, especially because manpower is stretched very thin already (there are two maintainers for this software, and going by stars alone that's a 400:1 user:developer ratio).
 
ckb-next relies on its community to provide [a PPA repository](https://launchpad.net/~davidmeikle/+archive/ubuntu/ckb-next-daily) of recent builds, and the extra manpower would be in my opinion better redirected towards updating that repository than merging our code into the Debuntu repositories. Alternatively, you could provide a third-party Debian-specific repository.
 
As more specific feedback to your commits:
- We would prefer the manual pages to be automatically generated if possible; it's going to be a maintenance burden to manually update them every time we add a command line option otherwise
- Typo fixes and removing the desktop encoding settings look good, thank you for catching these.
- As mentioned, we'd rather not have this in the Debian repositories, but we'll probably include your CMake flag to help third parties package this.

ZirconiumX:
888c10b and e49af76 have both been cherry-picked into master.

PR #377, opened 2019-03-02
Fix caps/num lock lag on K95 Platinum

gashton:
**System**: ArchLinux with Linux Kernel 4.20.13-arch1-1-ARCH

**Original Issue:** Animation updates would cause a significant lag in caps/num lock indicator updates.

**Workaround:** Locking the input mutex instead of the device mutex.

Tested for several months on K95 Platinum, no issues observed and resolves delay issue with caps/num lock indicators.

I did expect this to cause an issue as the USB device is accessed in function **os_sendindicators** in file **input_linux.c**. So I feel there may be a better solution, perhaps with priority locking?

### DIFF
--- a/src/daemon/input_linux.c
+++ b/src/daemon/input_linux.c
@@ -194,12 +194,12 @@ void* _ledthread(void* ctx){
                 ileds &= ~which;
         }
         // Update them if needed
-        pthread_mutex_lock(dmutex(kb));
+        pthread_mutex_lock(imutex(kb));
         if(kb->hw_ileds != ileds){
             kb->hw_ileds = ileds;
             kb->vtable->updateindicators(kb, 0);
         }
-        pthread_mutex_unlock(dmutex(kb));
+        pthread_mutex_unlock(imutex(kb));
     }
     return 0;
 }


tatokis:
Apologies for the late response, and thank you for the PR.

I don't think this is the right way of fixing it, as the `kb` object (usbdevice) is protected by a `dmutex`.
The input mutex protects other structures related to user input, and not anything inside `kb`.

Some time ago I started some work towards implementing a mutex specific to indicator LEDs, however it got complex really quickly and I eventually forgot about it.

I'm not really sure what should be done with this, as I don't think the correct action would be to merge this, even though it "fixes" the issue.

gashton:
Agreed, it sounds like another mutex to signal "next-to-access" would be a good idea. So RGB thread would lock ndmutex (next-to-access), then lock dmutex and unlock ndmutex before doing a small piece of work work. Then indicator thread could lock ndmutex, so that it gets to access next, and RGB thread would give up dmutex and wait for ndmutex to become available again. Effectively cheap round-robin scheduling algorithm, could be expanded to a priority based scheduling algorithm with an additional mutex too.

I might have a go at implementing this if I can find some time. 

But for now I'll close this pull request as it isn't the right solution. Thanks for looking into this though.

PR #400, opened 2019-04-18
Initial working snapcraft build

Kedstar99:
Hi Guys,

Sorry I haven't posted in a while, I just wasn't able to use my keyboard as I needed it at work.

I was looking at recent PR in the repo and if I have noticed that there are various PR so far relating to distributing ckb-next with Ubuntu and other derivatives. 

- Ubuntu doesn't update it's dependencies fast enough and this software is rolling so you don't want bug reports from people using old ckb-next versions.
- Building a PPA requires some launchpad to build the software with additional complications in the form of maintaining a changelist, control and the whole packing structure. It also requires interacting with bzr which is painful.
- From a user perspective, adding a PPA is kind of a bad idea as it gives root to the PPA owner and most users are unlikely to find, trust or use software distributed through this mechanism.
- It doesn't solve the issue of building for various release architectures.
- At the moment ckb-next seems to have several update mechanisms for install uninstall approaches that seem littered all over the place. There is codebase for handling various init, systemd scripts etc.

I have recently looked at ufw and how it's being distributed as a service for ubuntu and from that created this snapcraft mechanism. This approach should honestly make building for all the platforms that support snap, a lot easier.

- The build environment is frozen so you don't have to worry about package versions.
- It enables direct publication to the snapcraft store, or through their build service which is a lot simpler and doesn't require messing with bzr or launchpad.
- This repo can be added to a CI pipeline and it will directly handle build and distribution across several architectures. You will also get a nice worldmap showing who is using your software which may be nice.
- You won't need to worry about removal of old files as the snap will remove all remnants of the files in one location.
- You don't need to worry about services as it's handled by snapcraft and starts at startup. It also seems to work fine with interacting with my keyboard so i assume it works.
- You don't need to worry about updating as users are automatically pushed to the new version and can revert easily if necessary.
- It also comes sandboxed which some users may find comforting.
- It comes with all the non-dev and QT dependencies bundled which may be easier for users who don't want to install a bunch of dependencies they may not otherwise use.

Most importantly, it seems to work. I have tested a build for the snap and it interacts perfectly fine with my k70RGBlux. It remains in the background and all the colour interactions, keyboard presses seem to work. Audio controls work and windows key lock works. It correctly stays running in the background, on restart it all works.

Things to note/improve/test:
- The current approach declares the install prefix as /snap/ckb-next/current. It then moves the files to the correct location during the staging phase. 
- I haven't tested if macros work, but i can imagine it could hit an issue if it tries executing an external program.
- Theming support doesn't look the best, but it's being improved and I assume it will work in the future. I can make it look nice temporarily by say building it with say breeze/breeze-dark. I don't imagine it's the biggest issue for now though as i imagine most users won't be spending all their time in the program.


ZirconiumX:
I haven't read the patch or anything, but I will say this: welcome back.

Kedstar99:
> I haven't read the patch or anything, but I will say this: welcome back.

Thanks. I am happy to work on things when I can.  I can't do much long term as my company has banned my Razer keyboard at work for being to loud. As such, I have to use my corsair stuff at work which means I can't really use it for testing new ckb-next stuff.

This long Easter break gives me a little opportunity to do something as I have my Corsair keyboard at home.

Kedstar99:
Hi Peter,

I think I may have gotten this working because I was running on Kubuntu 19.04. Hence, it was able to pull everything from my local system.

I can fix this by using the desktop-launcher parts.

Sorry about this.

Kind regards,
Krish

peterzhang41:
Hi @Kedstar99 ,

No worries, I had a working version now. You can see the changes through revisions tab in the link. I just started learning snapcraft.

#snapcraft.yaml
https://github.com/ckb-next/ckb-next/pull/400#issuecomment-531153320

Kedstar99:
Hi @peterzhang41,

Would you mind submitting a PR to my repo. Then this PR gets updated with your commit and everyone can benefit. 

Kind regards,
Krish

peterzhang41:
PR is created.
https://github.com/Kedstar99/ckb-next/pull/1

Just curious, how did you use your local kubuntu as the build base. I can't find any info in the doc, and it keeps forcing me to build inside a virtual container (LXC, docker, or multipass)

Kedstar99:
When I was working on this, I don't think snapcraft required multipass for core18. As such, I just built it in a QEMU vm and it probably did stage stuff from the host os.

Things seemed to have changed since i last ran it.

Kedstar99:
Im just going to discard this, I don't think anyone wants to proceed and it doesn't seem needed tbh.

PR #411, opened 2019-05-16
Cope with kernels that do not have CONFIG_MODULES set

tonyvroon:
The current logic seems to assume the module loader is enabled, and that calling modprobe will always succeed. On a monolithic kernel without a module loader, that results in:
`May 16 09:20:42 agamemnon systemd[1]: Started Corsair Keyboards and Mice Daemon.
May 16 09:20:42 agamemnon ckb-next-daemon[17144]: ckb-next: Corsair RGB driver 0.4.0
May 16 09:20:42 agamemnon ckb-next-daemon[17144]: [I] Root controller ready at /dev/input/ckb0
May 16 09:20:47 agamemnon ckb-next-daemon[17144]: [I] Connecting Corsair Gaming K95 RGB PLATINUM Keyboard at /dev/input/ckb1
May 16 09:20:47 agamemnon ckb-next-daemon[17144]: modprobe: FATAL: Module uinput not found in directory /lib/modules/5.1.0-11058-g83f3ef3de625-dirty
May 16 09:20:47 agamemnon ckb-next-daemon[17144]: [F] os_inputopen (input_linux.c:62): Failed to load uinput module
May 16 09:20:47 agamemnon ckb-next-daemon[17144]: [I] Disconnecting /dev/input/ckb1
May 16 09:20:47 agamemnon ckb-next-daemon[17144]: [I] Removed device path /dev/input/ckb1
`

Changing the logic slightly helps it cope. Try to open the endpoints, if either of them work, we have the module and don't need to modprobe. If neither of them work *and* the modprobe fails, that's cause for a fatal error.

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -281,15 +281,6 @@ void MainWindow::updateVersion(){
         QString kextstatOut(kextstat.readAll());
         if(kextstatOut.isEmpty())
             daemonWarning.append(tr("<br /><b>Warning:</b> System Extension by \"Fumihiko Takayama\" is not allowed in Security & Privacy. Please allow it and then unplug and replug your devices."));
-#elif defined(Q_OS_LINUX)
-            QProcess modprobe;
-            modprobe.start("modprobe", QStringList("uinput"));
-
-            if(!modprobe.waitForFinished())
-                qDebug() << "Modprobe error";
-
-            if(modprobe.exitCode())
-                daemonWarning.append(tr("<br /><b>Warning:</b> The uinput module could not be loaded. If this issue persists after rebooting, compile a kernel with CONFIG_INPUT_UINPUT=y."));
 #endif
         settingsWidget->setStatus(tr("No devices connected") + daemonWarning);
     }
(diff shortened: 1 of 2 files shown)


ZirconiumX:
You'll also need to patch the GUI so that it doesn't pop up with a warning
about uinput being unavailable.

tonyvroon:
Thank you to dwfreed on #gentoo-dev for code review; I had an extraneous close in my error path.

ZirconiumX:
...But *not* by removing it.

Remember that we're trying to be user-friendly. We have people who build CKB by just running the `quickinstall` script and not knowing how it works under the hood; if there's a kernel problem that somebody doesn't realise, we'll get complaints that "it doesn't work" without the user knowing the reason for it.

tonyvroon:
I'm happy to take direction here; what should the UI do instead?

ZirconiumX:
Duplicate the logic you put in the daemon; it's a relatively small maintenance overhead to have two copies of it.

src/gui/mainwindow.cpp
```diff
@@ -282,14 +282,20 @@ void MainWindow::updateVersion(){
...
-
-            if(!modprobe.waitForFinished())
-                qDebug() << "Modprobe error";
-
-            if(modprobe.exitCode())
-                daemonWarning.append(tr("<br /><b>Warning:</b> The uinput module could not be loaded. If this issue persists after rebooting, compile a kernel with CONFIG_INPUT_UINPUT=y."));
+            QFileInfo uinput("/dev/uinput");
+            if (!uinput.exists()){
+                QFileInfo uinput2("/dev/input/uinput");
+                if (!uinput2.exists()){
```
ZirconiumX: This logic feels clumsy to me. I'm more inclined to have these be something like
```cpp
if (!QFileInfo("/dev/uinput").exists())
```
but ultimately @tatokis has the final word on things involving the GUI.

tatokis:
Merged in https://github.com/ckb-next/ckb-next/commit/7e7f219ef8538d5df631e56ac0d4a1678244254d

PR #422, opened 2019-05-29
Do not prepend /usr to libexecdir

### LINKED ISSUE #276: Can't find the animation directory after latest update.
I have not tried to build from source.. I'm unsure of what will happen since I'm using the copr build. And mentioning that.. I'm not sure where to correct post this issue..

### Source

copr build on F28

## What is the program's current behavior?

The <insert name> script could not be loaded. Please check your animation directory.

## What is the expected behavior of the program?

Loads animations correctly.

## Any other notes & comments?

---

# Something else

bmwiedemann:
Fix #276

Without this patch, compiling with -DCMAKE_INSTALL_LIBEXECDIR=%{_libexecdir}
expanded this to /usr//usr/lib/ckb-next-animations

and there is no rpm macro to have libexecdir without /usr
but we need to use these macros to be able to use the same .spec file
across distributions (Fedora has /usr/libexec and openSUSE has /usr/lib )

This matches usage in src/daemon/CMakeLists.txt

Also tracked at https://bugzilla.opensuse.org/show_bug.cgi?id=1135528


Thaodan:
If I understand everything correctly the GNUInstallDirs cmake module sets LIBEXECDIR to libexec and the ckb-next expects this and appends ${PREFIX} to it right?
Usally it works like this when you override libexecdir to make it to $libdir/appname because the distro doesn't use libexec, the buildsystem doesn't append anything to it.
So the commit is right but there's need for another one to workaround that GNUInstallDirs doesn't add the ${PREFIX} (which is a feature btw).

I think there needs to be something like:
`set(CMAKE_INSTALL_LIBEXECDIR "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBEXECDIR}" STRING Cache)`

tatokis:
Closing due to inactivity. If this issue still exists, feel free to open another PR with a fix that addresses the issues I raised above.

PR #460, opened 2019-08-29
Fixed mouse SYN

Sur3:
Mouse SYN for X and Y pair are now shared also isync is seperated into isync_mouse and isync_keyboard.

### DIFF
--- a/src/daemon/input_linux.c
+++ b/src/daemon/input_linux.c
@@ -170,17 +170,21 @@ void os_mousemove(usbdevice* kb, int x, int y){
     event.type = EV_REL;
     int fd=kb->uinput_mouse - 1;
     //send X
-     event.code = REL_X;
-     event.value = x;
-     if(write(fd, &event, sizeof(event)) <= 0)
-         ckb_warn("uinput write failed: %s\n", strerror(errno));
+    if(x!=0){
+        event.code = REL_X;
+        event.value = x;
+        if(write(fd, &event, sizeof(event)) <= 0)
+            ckb_warn("uinput write failed: %s\n", strerror(errno));
+    }
     //send Y
-     event.code = REL_Y;
-     event.value = y;
-     if(write(fd, &event, sizeof(event)) <= 0)
-         ckb_warn("uinput write failed: %s\n", strerror(errno));
+    if(y!=0){
+        event.code = REL_Y;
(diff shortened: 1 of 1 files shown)


src/daemon/input_linux.c
```diff
@@ -128,14 +128,22 @@ void os_inputclose(usbdevice* kb){
     kb->uinput_mouse = 0;
 }
 
-// Generate SYN reports to synchronize device
-static void isync(usbdevice* kb){
+// Generate SYN reports to synchronize keyboard device
+static void isync_kb(usbdevice* kb){
```
tatokis: I'm not sure it's a good idea to separate them when they essentially do the same thing. Perhaps instead of passing a usbdevice* you can pass an int fd to the function and keep only one copy of the code that does both.

src/daemon/input_linux.c
```diff
@@ -159,32 +167,35 @@ void os_keypress(usbdevice* kb, int scancode, int down){
         event.value = down;
         is_mouse = !!(scancode & SCAN_MOUSE);
     }
-    if(write((is_mouse ? kb->uinput_mouse : kb->uinput_kb) - 1, &event, sizeof(event)) <= 0)
-        ckb_warn("uinput write failed: %s\n", strerror(errno));
-    else
-        isync(kb);
+    if(is_mouse){
```
tatokis: This can be simplified by having a single write call and picking the appropriate handle based on the if.
For example

```C
int fd = (is_mouse ? kb->uinput_mouse : kb->uinput_kb) - 1;
// [...]
if(write(fd, [...]
```

Using the same thing you can also use a unified isync (as mentioned above) and just call isync unconditionally in the end with the fd passed to it.

Sur3:
@tatokis Thanks for the review, I incorporated your suggested changes.
I just wonder if it were probably better to use a single usbdevice as parameter instead of the file descriptor as using the file descriptor breaks a bit with the abstract model of object orientation. But that's only cosmetics of course.

Sur3:
I also added back the if conditions around sending the mouse coordinates, because even though zero coordinates get ruled out by evdev anyway a write call is probably more overhead than checking the condition. Only statistics about how often one coordinate is zero could tell if having the additional if statement is better, but as we have a high poll-rate that could be the case more often.

PR #471, opened 2019-09-17
Harpoon RGB PRO support

Turtleb01:
Quick patch, adds Pro model id to IS_HARPOON.
I don't see any differences between PRO and non-PRO other than sensor (6000dpi vs 12000dpi).
No extensive testing done; works on my mouse and shouldn't break anything.
Firmware update thing should maybe be also updated
[ckb-next-dev-detect](https://turtleb01.github.io/ckb-next-dev-detect.txt) if somebody needs this
I'm new to Github, sorry if I'm doing something wrong.


tatokis:
Please use git rebase instead of merging and then git push --force to this branch.

Turtleb01:
I really don't know how Github works, just take those 7 lines of "code" :D

PR #517, opened 2019-12-21
experiment with temporal dithering

m9710797:
Hi,

First of all, do NOT actually merge this pull request! It breaks existing functionality! I merely want to use this as a starting point for a discussion.

Problem description:
I recently bought a Corsair Ironclaw mouse. I wanted to give the logo-LED a slow/dim glowing animation (black -> dark red -> black, over the course of several seconds). Unfortunately I found that the hardware isn't well suited to display dark colors. There is a too big difference in brightness between black (0,0,0) and the darkest red (1,0,0). Also the next steps are too big (1,0,0)->(2,0,0)->(3,0,0)->... (OTOH there's hardly any noticeable difference between e.g. (250,0,0) and (255,0,0)).

To solve this I experimented with temporal dithering. With the (hacked) code in this PR I am able to get a pretty nice looking animation. The next step is to transform this hack into a more general solution. My question is "are you interested in contributions for such a feature?". In other words: Should I simply change the code so that it's good enough for me? Or should I immediately try to code it in a form that's acceptable to upstream?

There are two main problems that should be overcome:
1) Currently colors are represented by 3x8-bit RGB triplets. But (for dark colors) this representation isn't accurate enough. My first idea is to switch the color components from 'uchar' to 'float'. And only at the end points (QRgb objects for the GUI / the USB data packets for the hardware) convert these back to integers.

2) To avoid flickering the effect needs to be run at a high enough frame rate. For me 120Hz gave an acceptable result. Though currently the software already gives a warning for rates above 30Hz. I'd like to understand the reason for this warning. Is it only about high CPU usage, or is it more serious? At 120Hz I did indeed see a noticeable higher CPU usage (I think there are ways to improve upon this, but let's take one step at a time).

Wouter

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -146,7 +146,8 @@ if (MACOS OR LINUX)
         ckb-next-daemon
             PRIVATE
               Threads::Threads
-              "${ICONV_LIBRARIES}")
+              "${ICONV_LIBRARIES}"
+              m)
 endif ()
 if (MACOS)
     target_link_libraries(
--- a/src/daemon/led_mouse.c
+++ b/src/daemon/led_mouse.c
@@ -3,9 +3,34 @@
 #include "profile.h"
 #include "usb.h"
 
+#include <math.h>
+
 // Compare two light structures, ignore keys
 static int rgbcmp(const lighting* lhs, const lighting* rhs){
     return memcmp(lhs->r + LED_MOUSE, rhs->r + LED_MOUSE, N_MOUSE_ZONES) || memcmp(lhs->g + LED_MOUSE, rhs->g + LED_MOUSE, N_MOUSE_ZONES) || memcmp(lhs->b + LED_MOUSE, rhs->b + LED_MOUSE, N_MOUSE_ZONES);
+}
+
+// uchar [0 .. 255] -> float [0.0 .. 1.0]
+static float toFloat(uchar x)
+{
+    return (float)x / 255.0f;
+}
+
+// for now assume 256 steps
+static float quantize(float x)
+{
+    return roundf(x * 255.0f) / 255.0f;;
+}
+
+// float [0.0 .. 1.0] -> uchar [0 .. 255]
+//  Values outside [0.0 .. 1.0] are clipped
+//  (But for dithering this shouldn't happen.)
+static uchar toUchar(float x)
+{
+    int i = (int)(x * 255.0f);
+    if (i < 0) return 0;
+    if (i > 255) return 255;
+    return i;
 }
 
 int updatergb_mouse(usbdevice* kb, int force){
@@ -17,7 +42,7 @@ int updatergb_mouse(usbdevice* kb, int force){
     if(!force && !lastlight->forceupdate && !newlight->forceupdate
             && !rgbcmp(lastlight, newlight))
         return 0;
-    lastlight->forceupdate = newlight->forceupdate = 0;
+    lastlight->forceupdate = newlight->forceupdate = 1;
 
     // Prevent writing to DPI LEDs or non-existent LED zones for the Glaive.
     int num_zones = IS_GLAIVE(kb) ? 3 : N_MOUSE_ZONES;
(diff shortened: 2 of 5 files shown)


ZirconiumX:
I'd like to apologise for the delay on getting back to you on this; I've kinda burned out on ckb.

> Currently colors are represented by 3x8-bit RGB triplets. But (for dark colors) this representation isn't accurate enough. My first idea is to switch the color components from 'uchar' to 'float'. And only at the end points (QRgb objects for the GUI / the USB data packets for the hardware) convert these back to integers.

Round-tripping from float to string to float is a kind of painful solution; parsing floats from strings is slow and bug-prone, floating-point math is slow and bug-prone and I'm not sure I trust even myself to get float parsing right considering locale issues such as the decimal separator. I'd prefer expanding the bit width to something like short or int instead.

> To avoid flickering the effect needs to be run at a high enough frame rate. For me 120Hz gave an acceptable result. Though currently the software already gives a warning for rates above 30Hz. I'd like to understand the reason for this warning. Is it only about high CPU usage, or is it more serious? At 120Hz I did indeed see a noticeable higher CPU usage (I think there are ways to improve upon this, but let's take one step at a time).

Indeed, CPU usage rises in correlation with animation frequency, because we're effectively software rendering to the mouse, which also results in increased USB traffic.

I think temporal dithering is an interesting idea, and might be of relevance to the older 9-bit protocol keyboards, so I'd be inclined to merge it if it was of suitable code quality.

m9710797:
> I'd like to apologise for the delay on getting back to you on this; I've kinda burned out on ckb.

No problem at all.

> Round-tripping from float to string to float is a kind of painful solution; parsing floats from strings is slow and bug-prone, floating-point math is slow and bug-prone and .... I'd prefer expanding the bit width to something like short or int instead.

I only partly agree. Converting float->string->float is indeed slower than int->string->int. But floating point math is not slower than integer math (on Intel/ARM floating point addition/multiplication instructions take the same number of cycles as integer addition/multiplication). In fact floating point could even be faster (and simpler): e.g. scaling a floating point is just a multiplication, while scaling an integer is often implemented as a multiplication followed by a shift (e.g. to get the result back in the 0..255 range). The latter is slower and, in some cases, prone to overflow problems.

Anyway, speed of the calculation is not an issue. The conversion to and from string is. But also the inter-process communication (passing through the kernel) has a significant cost.

Ideally I would like to remove most inter-process communication. That means the daemon process itself must have the _standard_ animations builtin (there can still be _extra_ external animations that continue to use IPC). And the daemon process itself knows how to combine those animations into the final result. In such a scenario there is no need anymore for conversions to/from string.

I do realize this is quite a radical change and quite a lot of work. And maybe I am overlooking something fundamental. So before writing any code, I think it's important to thoroughly discuss this.

> Indeed, CPU usage rises in correlation with animation frequency, because we're effectively software rendering to the mouse, which also results in increased USB traffic.

I think the CPU usage part is solvable (and could even be improved compared to today's situation). But of course you're right about the USB traffic.

> I think temporal dithering is an interesting idea, and might be of relevance to the older 9-bit protocol keyboards, so I'd be inclined to merge it if it was of suitable code quality.

How do you see this going forward? I'm afraid that if I simply implement (all) my idea(s), the result might not be acceptable to you. So I'd first like to agree on the general direction and then think of a plan for how to split that work in small steps. Are you willing to help me with that?

Thanks.

ZirconiumX:
> In fact floating point could even be faster (and simpler): e.g. scaling a floating point is just a multiplication, while scaling an integer is often implemented as a multiplication followed by a shift (e.g. to get the result back in the 0..255 range). The latter is slower and, in some cases, prone to overflow problems.

As somebody who works with FPGAs as a hobby/career, I shudder whenever somebody says "just a multiplication" 😛.

> Ideally I would like to remove most inter-process communication. That means the daemon process itself must have the _standard_ animations builtin (there can still be _extra_ external animations that continue to use IPC). And the daemon process itself knows how to combine those animations into the final result. In such a scenario there is no need anymore for conversions to/from string.

@tatokis has in the past vetoed a suggestion of this (from me) because he thinks animations belong in the GUI, not the daemon, and that does correlate with the Principle of Least Privilege. While I do agree with you here - especially with devices with hardware animation support in mind - without a consensus between us, that idea won't get very far.

> I do realize this is quite a radical change and quite a lot of work. And maybe I am overlooking something fundamental. So before writing any code, I think it's important to thoroughly discuss this.

Highly agreed.

> I think the CPU usage part is solvable (and could even be improved compared to today's situation). But of course you're right about the USB traffic.

If you find something, feel free to submit a PR.

> How do you see this going forward? I'm afraid that if I simply implement (all) my idea(s), the result might not be acceptable to you. So I'd first like to agree on the general direction and then think of a plan for how to split that work in small steps. Are you willing to help me with that?

The most immediate thing that comes to mind is that ckb [already has](https://github.com/ckb-next/ckb-next/blob/master/src/daemon/led_keyboard.c#L48) a form of dithering for 9-bit keyboards. Suppose we extend the colour-space (somehow), then we could have the dithering functions apply on both 9-bit and 24-bit keyboards, which may be anything from quantisation to your temporal dithering.

m9710797:
> As somebody who works with FPGAs as a hobby/career, I shudder whenever somebody says "just a multiplication" stuck_out_tongue.

Hehe, I understand. Personally I'm more a software oriented person, but I work closely together with hardware people (we develop tools to design ASIPs).

> @tatokis has in the past vetoed a suggestion of this (from me) because he thinks animations belong in the GUI, not the daemon, and that does correlate with the Principle of Least Privilege. While I do agree with you here - especially with devices with hardware animation support in mind - without a consensus between us, that idea won't get very far.

I'm very interested in hardware animation support. What's the status of reverse engineering that part? Is there anything I could do to help with that? E.g. the firmware is ARM code, right? Can it easily be disassembled or is it encrypted in some way? Maybe I could help by annotating that ARM code?


> > I do realize this is quite a radical change and quite a lot of work. And maybe I am overlooking something fundamental. So before writing any code, I think it's important to thoroughly discuss this.

> Highly agreed.

Ok, here are some initial steps for a plan (far from complete):

Requirements:
* Current color representation is 24-bit RGB. This is (for me) not sufficient for dark colors.
* Current maximum frame-rate is 30fps. This is insufficient for temporal dithering. Increasing it to 120Hz seems to work fine, but increases CPU usage (and USB traffic). So anything we can do to reduce CPU usage is good. (Also, ckb-next is running all the time on my computer, so even at 30Hz it's good to get CPU usage as low as reasonably possible).

Color representation, current situation:
* Individual animations: currently these produce 24-bit RGB. I can live with that: if I want a darker output I can work around this by e.g. lowering the opacity in the animation script. OTOH internally many animations actually calculate the colors using float, and only round to integers before outputting the result.
* Full animation script (src/gui/kbanim.cpp): this takes two colors represented as QRgb (3x8-bit integer). It converts the RGB components to float and applies some combination-formula. This result is then converted back to QRgb.
* ckb-daemon: currently this one receives 24-bit rgb colors. In my current hack I convert this representation to float and then perform the temporal dithering calculation. But wider integers would work equally well.

So in a way most of the color-pipeline is already float based. Only at the communication boundaries the values are rounded to int (and converted to/from string). Because of this I thought it would be a natural transformation to remove the int<->float conversions. Alternatively choosing some wider integer based representation is  also possible. But that would then be a 3rd color representation:
1) float for (most) calculations
2) wide-integer at the communication boundaries
3) 24-bit integer to talk to the hardware.

Backwards compatibility:
I'm not very familiar with the history of ckb(-next) and with all the possible use cases. How do you typically handle upgrades? Is it a requirements that the gui, the daemon and the animations are upgraded together or is the user allowed to mix versions? And specifically external animations. Should those continue to run with newer ckb versions? Or is it acceptable to require a recompile?

PR #547, opened 2020-02-22
Replace deprecated QString::sprintf with a string

dkorkmazturk:
I was just wondering if usage of QString::sprintf is really needed here. The same file contains a call to cmd.write() function that does not use QString::sprintf, just like in my change.

If I am missing something or toLatin1() is needed, I think cmd.write(QString("rgb 000000").toLatin1()); can be used instead.


tatokis:
It's probably a leftover. Thanks!

There's another (actual) use of sprintf in there, but until it gets completely removed, I don't think I am messing with it. We could replace it with asprintf instead, but that's also not recommended and it was introduced in Qt5.5. It's not obsolete though.

dkorkmazturk:
Oh I missed the other sprintf. I've just created https://github.com/ckb-next/ckb-next/pull/549 for that one.

PR #549, opened 2020-03-01
Replace deprecated QString::sprintf with QTextStream

dkorkmazturk:
I overlooked one of the QString::sprintf() usages remaining in this file. Since QString::asprintf is not recommended either, I decided to use QTextStream to replace the last QString::sprintf in this file.

### DIFF
--- a/src/gui/ckbsettings.cpp
+++ b/src/gui/ckbsettings.cpp
@@ -56,7 +56,11 @@ static QSettings* globalSettings(){
 }
 
 bool CkbSettings::isBusy(){
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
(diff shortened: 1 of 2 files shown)


tatokis:
So, what do we do with this?

As mentioned, the rest of the code uses C sprintf, which according to some quick tests I did is much faster than using QString::sprintf.
QTextStream is slightly faster than QString::sprintf, but much slower than C sprintf.

If we go with performance, it would need to be converted to C sprintf with a char[].
If we go with readability, QTextStream is really difficult to read compared to a format string.

I really do not know what the right thing to do is, here.
Other than replacing something that's deprecated, I see very little actual advantage to using QTextStream to justify a change. But it's deprecated, sooo...

PR #553, opened 2020-03-13
Performance tidying

dkorkmazturk:
I decided to run some clang-tidy checks on ckb-next's gui directory and let it fix the reported problems. These commits contain fixes for the results of following clang-tidy checks:

- performance-unnecessary-value-param

- performance-unnecessary-copy-initialization

- performance-type-promotion-in-math-fn

### DIFF
--- a/src/gui/animscript.cpp
+++ b/src/gui/animscript.cpp
@@ -194,7 +194,7 @@ bool AnimScript::load(){
             QMap<QString, QVariant> preset;
             for(int i = 2; i < count; i++){
                 // Scan name/value setting pairs
-                QString setting = components.at(i);
+                const QString& setting = components.at(i);
                 QStringList sComponents = setting.split("=");
                 if(sComponents.count() != 2)
                     continue;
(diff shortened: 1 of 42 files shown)


ZirconiumX:
Finally got round to looking at this. If we're going for performance improvements, it doesn't make much sense to use the `double` versions of floating-point functions instead of the plain `float` versions.

dkorkmazturk:
These functions are overloaded in C++ and compilers detect correct versions of them so there is no need to explicitly giving the correct versions of these functions. So, if the argument is float, float version will be called and if the argument is double, double version will be called.

As we discussed on IRC, I know you guys prefer using C functions that are dedicated to a certain type (Like fabsf for float, fabs for double etc.), but I still think that these files are written in C++ and they should follow modern C++ rules.

tatokis:
I was initially going to cherry pick the two commits and make a new one to use the C-like functions, but in the end I just cherry picked the third one as well, so I basically merged it.

PR #554, opened 2020-03-15
Allow manual override of systemd unit directory

ryanfitzsimon:
Add a CMake variable "SYSTEMD_UNIT_INSTALL_DIR" that can be manually set
to change where unit files are installed.

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -304,6 +304,7 @@ elseif (LINUX)
     if (EXISTS "/run/systemd/system" OR "systemd" IN_LIST FORCE_INIT_SYSTEM)
         message(STATUS "systemd detected")
         set(CKB_NEXT_INIT_SYSTEM "systemd" CACHE INTERNAL "")
+        set(SYSTEMD_UNIT_INSTALL_DIR "/usr/lib/systemd/system" CACHE STRING "Where to install systemd unit files.")
         set(DISALLOW_SYSVINIT TRUE)
 
         # Generate and import service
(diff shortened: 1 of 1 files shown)


ryanfitzsimon:
This doesn't entirely resolve the problems with differing unit file directories across distros discussed in #273, but at least it makes it possible to use without patching when packaging for systems where `/usr/lib/system/systemd` won't work.

ryanfitzsimon:
Always installing to `/usr/lib/systemd/system` is not strictly correct, but thanks to symlinks we can get away with it for distros in the table below.

| OS       | System Unit File Directory | What is /lib?    |
|----------|----------------------------|------------------|
| Arch     | `/usr/lib/systemd/system`  | link to `/usr/lib` |
| Debian   | `/lib/systemd/system`      | link to `/usr/lib` |
| Fedora   | `/lib/systemd/system`      | link to `/usr/lib` |
| OpenSuse | `/usr/lib/systemd/system`  | directory        |
| Ubuntu   | `/lib/systemd/system`      | directory        |

#273 attempted to use `CMAKE_INSTALL_LIBDIR` to construct the unit file directory, but this doesn't work for all systems (it was reverted in 6306c688e96c69a46076ad6d64a4dfb3641e8167).
For example, Fedora sets `CMAKE_INSTALL_LIBDIR` to `/usr/lib64/`which is a real directory (not a symlink to `/usr/lib` like on Arch) and therefore not searched by systemd.

It's left up to package maintainers to correctly set `SYSTEMD_UNIT_INSTALL_DIR` in order to install to the official system unit file location.

tatokis:
Thank you!

PR #561, opened 2020-04-10
WIP: Allow to select if mute indicator is based on standard sink or source (microphone)

penguineer:
I wanted to change the mute indicator to show my microphone state (alongside the change in my window manager to mute the microphone when I press the mute button). With this change the source can be configured in the keyboard profile.

Unfortunately I cannot test this on a MAC, so the feature is just unavailable there.

I leave it open for discussion if the source setting should be based on the profile or in the general settings.

### DIFF
--- a/src/gui/kperfwidget.cpp
+++ b/src/gui/kperfwidget.cpp
@@ -57,6 +57,15 @@ KPerfWidget::KPerfWidget(QWidget *parent) :
             });
         }
     }
+    muteDev = ui->muteDev;
+    if (isMuteDeviceSupported())  {
+        muteDev->setEnabled(true);
+        connect(muteDev, OVERLOAD_PTR(int, QComboBox, currentIndexChanged), [=] (int index) {
+            perf->setMuteDevice(static_cast<muteDevice>(index));
+        });
+    } else
+        muteDev->setEnabled(false);
+
     k95Widgets << ui->modeBox << ui->modeColorOn << ui->modeColorOff << ui->macroBox << ui->macroColorOn << ui->macroColorOff << ui->k95Label1 << ui->k95Label2 << ui->k95Label3 << ui->k95Label4 << ui->k95Label5 << ui->k95Label6 << ui->k95Line << ui->k95Spacer;
 }
 
@@ -170,6 +179,9 @@ void KPerfWidget::setPerf(KbPerf* newPerf, KbProfile* newProfile){
             if(indicators[i].color3) indicators[i].color3->setEnabled(false);
         }
     }
+    // Set mute device
+    if (muteDev)
+        muteDev->setCurrentIndex(perf->getMuteDevice());
     // Hide K95 indicators on non-K95s
     if(profile->keyMap().model() == KeyMap::K95){
         foreach(QWidget* w, k95Widgets)
--- a/src/gui/kperfwidget.ui
+++ b/src/gui/kperfwidget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>840</width>
-    <height>679</height>
+    <width>843</width>
+    <height>737</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -26,14 +26,44 @@
    <property name="bottomMargin">
     <number>9</number>
    </property>
-   <item row="22" column="5">
-    <widget class="QLabel" name="label_20">
+   <item row="3" column="1" colspan="2">
+    <widget class="QLabel" name="label_2">
      <property name="text">
-      <string>On</string>
+      <string>Indicator intensity:</string>
      </property>
     </widget>
    </item>
-   <item row="23" column="4">
+   <item row="16" column="9">
+    <spacer name="horizontalSpacer_6">
+     <property name="orientation">
+      <enum>Qt::Horizontal</enum>
+     </property>
+     <property name="sizeType">
+      <enum>QSizePolicy::Fixed</enum>
+     </property>
+     <property name="sizeHint" stdset="0">
+      <size>
+       <width>10</width>
+       <height>20</height>
+      </size>
+     </property>
+    </spacer>
+   </item>
+   <item row="21" column="8">
+    <widget class="QLabel" name="label_23">
+     <property name="text">
+      <string>Off</string>
+     </property>
+    </widget>
+   </item>
+   <item row="22" column="1">
+    <widget class="QLabel" name="label_7">
+     <property name="text">
+      <string notr="true">Scroll lock:</string>
+     </property>
+    </widget>
+   </item>
+   <item row="22" column="4">
     <widget class="ColorButton" name="scrollColorOn">
      <property name="sizePolicy">
       <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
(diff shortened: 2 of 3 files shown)


penguineer:
I wonder if parallel to the indicator source the mute button could also be rewired on the ckb-next side to either mute the microphone or the speakers. Currently I do this in my window manager.

ZirconiumX:
Since it's no longer WIP would you like us to review it?

src/gui/kbperf.cpp
```diff
@@ -132,6 +133,10 @@ void KbPerf::load(CkbSettings& settings){
         }
     }
     SGroup group(settings, "Performance");
+    {
+        QString s_muteDev = settings.value("MuteDevice", "sink").toString();
+        iMuteDev = s_muteDev == "source" ? SOURCE : SINK;
```
tatokis, line 138: Please use parentheses around ternary operators.
Makes them easier to read.

src/gui/kbperf.cpp
```diff
@@ -211,6 +216,7 @@ void KbPerf::load(CkbSettings& settings){
 void KbPerf::save(CkbSettings& settings){
     _needsSave = false;
     SGroup group(settings, "Performance");
+    settings.setValue("MuteDevice", iMuteDev == SOURCE ? "source" : "sink");
```
tatokis, line 219: Same as above

src/gui/kperfwidget.cpp
```diff
@@ -57,6 +57,16 @@ KPerfWidget::KPerfWidget(QWidget *parent) :
             });
         }
     }
+    muteDev = ui->muteDev;
+//    muteDev->setCurrentIndex(perf ? perf->getMuteDevice() : 1);
```
ZirconiumX, line 61: Can you clean this comment up please?

src/gui/kperfwidget.cpp
```diff
@@ -57,6 +57,16 @@ KPerfWidget::KPerfWidget(QWidget *parent) :
...
+    if (isMuteDeviceSupported())  {
+        muteDev->setEnabled(true);
+        connect(muteDev, QOverload<int>::of(&QComboBox::currentIndexChanged), [=] (int index) {
```
tatokis, line 64: We have a Q_OVERLOAD macro, please use that as it retains compatibility with older Qt versions that we still support.

src/gui/kperfwidget.ui
```diff
@@ -94,6 +94,20 @@
      </property>
     </widget>
    </item>
+   <item row="18" column="4" colspan="8">
+    <widget class="QComboBox" name="muteDev">
+     <item>
+      <property name="text">
+       <string>Use default sink (Audio)</string>
```
tatokis, line 101: These are not user-friendly names.
I'd suggest "Use default output device"

src/gui/kperfwidget.ui
```diff
@@ -94,6 +94,20 @@
...
+      </property>
+     </item>
+     <item>
+       <property name="text">
+        <string>Use default source (Microphone)</string>
```
tatokis, line 106: Possibly "Use default input device"

src/gui/media_linux.cpp
```diff
@@ -33,16 +35,26 @@ static void SinkCallback(pa_context* context, const pa_sink_info* info, int eol,
...
+    lastKnownSink = info->mute ? MUTED : UNMUTED;
+}
+
+static void SourceCallback(pa_context* context, const pa_source_info* info, int eol, void* data){
+    QMutexLocker locker(&mutex);
+
+    if(info == nullptr || info->name != defaultSource)
+        return;
+
+    lastKnownSource = info->mute ? MUTED : UNMUTED;
```
tatokis, line 47: Parentheses around here would be appreciated

src/gui/media_linux.cpp
```diff
@@ -102,7 +124,15 @@ static void ContextStateCallback(pa_context* context, void* data){
 
 #endif
 
-muteState getMuteState(){
+bool isMuteDeviceSupported() {
+    return true;
+}
+
+muteState mute_state(const muteDevice muteDev) {
+    return muteDev == SINK ? lastKnownSink : lastKnownSource;
```
tatokis, line 132: Parentheses

src/gui/kperfwidget.cpp
tatokis: Sorry, it's called `OVERLOAD_PTR` https://github.com/ckb-next/ckb-next/blob/050231f2560fdfa25668cea1a57301dc79b944d4/src/gui/qoverloadlegacy.h

penguineer:
Thank you for the feedback! I will check and push the changes.

src/gui/kbperf.cpp
penguineer: I understand that you would like to see `(s_muteDev == "source") ? SOURCE : SINK`

penguineer:
I have fixed the issues. Please let me know if you would like them  squashed into the original commits.

penguineer:
It seems that the setting is not copied when the profile is duplicated. I need to fix this.

penguineer:
What can I do to help with this PR?

tatokis:
Sorry, I've been busy with other things. 
I didn't like how the ui was laid out and I was faking to suggest some changes that required changing other parts of the gui well. 

I made a mockup which I'll share later tonight once I find it.

penguineer:
> Sorry, I've been busy with other things. I didn't like how the ui was laid out and I was faking to suggest some changes that required changing other parts of the gui well. I made a mockup which I'll share later tonight once I find it.

Okay, thank you for the quick reply!

I am open  to suggestions for the UI.

tatokis:
I was thinking something like this
7cd6aa4a2edb

penguineer:
If I understand your mock-up correctly, there is a decision to be made:

A) Use your proposal, which is visually more appealing, because it brings more order to the way the dialog is laid out.
B) Use my proposal which adheres to the UX principle to keep similar functions (configuration for the mute button) close together.

Which way do we go?

tatokis:
A month later... (Sorry.)
After thinking about it, you are right. Unfortunately both aren't great solutions.
I do prefer my suggestion, as it doesn't break layout, as I tend to get annoyed when I see bad layouts, but it isn't intuitive.

I do think that consistent layout was the thought process when the UI was originally designed by ccMSC, as the "indicator intensity" option wasn't placed under "Top row" but rather under Miscellaneous, on its own.

penguineer:
Okay, I will change it to your suggestion then.

In case we have more discussion about the UI layout, I would liker to put the "backend" part, which we agree on, in a distinct PR to get it merged. Would that be an option for you?

tatokis:
I'm not sure I understand correctly. Wouldn't doing that add dead code?

penguineer:
> I'm not sure I understand correctly. Wouldn't doing that add dead code?

More like "dormant" code. It still extends the code base and  gives others the chance to use or extend it. But it means that the back-end part can already be taken into account during changes while we figure out the front-end.

tatokis:
Sure, go ahead

penguineer:
> Sure, go ahead

Thank you, I've submitted #661

penguineer:
I created a different layout based on your suggestion. Please have a look. :)

tatokis:
So, apologies for this (as it was my suggestion). I just had a look and I feel that the "copy to mode" button is a bit confusing there.

Perhaps it would be better if it was moved below "Indicate mute from", and renamed to something like "Copy performance settings to mode..." or something along those lines.

tatokis:
https://github.com/ckb-next/ckb-next/commit/ca7c2ba010012f1d8c98cf084efb405369e05aa9

PR #583, opened 2020-05-13
Fix gcc-10/-fno-common error

zeekec:
gcc-10 now defaults to -fno-common and does not automatically add extern
to variables declared in headers.

https://gcc.gnu.org/gcc-10/porting_to.html

Signed-off-by: Erik Zeek <zeekec@gmail.com>


ZirconiumX:
Thanks for catching this!

PR #603, opened 2020-06-05
add point judge

yuantaiyong:
add  arg  of  point  NULL  judge


tatokis:
What does this PR exactly do?

The function is only called here, and &indev can not be null.
https://github.com/ckb-next/ckb-next/blob/7fc4be1d5811686b55cad36ffed0524e7bc3077d/src/daemon/input_linux.c#L93-L102

Additionally, if there was ever an issue, it would show up in the `memset(&indev,...)` which is called immediately after indev is created.

Issue #592, opened 2020-05-21
Add JOBS env var for quickinstall


tatokis:
It might be helpful in the future to make quickinstall check if JOBS is set before running it, and have it override the auto detection.

Something along the lines of
```
JOBS=2 ./quickinstall
```
should result in
```
[...] make -j 2 [..]
```

audreyseo:
This should be closed, right?

tatokis:
Yup!

PR #612, opened 2020-06-25
Add JOBS env var for quickinstall

neeldug:
Resolves issue #592 by using parameter expansion that resolves when JOBS is explicitly set as an environment variable.

### DIFF
--- a/quickinstall
+++ b/quickinstall
@@ -52,7 +52,9 @@ cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Release -DSAFE_INSTALL=ON -DSAFE_UNINSTALL=
 checkfail $?
 
 # Find out how many cores the system has, for make
-JOBS=$(getconf _NPROCESSORS_ONLN 2>/dev/null)
+if [ -z ${JOBS+var} ]; then
+	JOBS=$(getconf _NPROCESSORS_ONLN 2>/dev/null)
+fi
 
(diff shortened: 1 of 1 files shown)


tatokis:
How does this work exactly?
I'm not entirely sure I understand how it works, and why you are using `var` in this case, and not something else.

Unless I'm mistaken, can't the same be achieved with another `if [[ -z "$JOBS" ]]; then`?
I think that's more easily understandable.

neeldug:
Yeah sorry about the confusion, the var is only needed if JOBS was set to an empty string by accident, it’s essentially there to prevent erroring if JOBS was set at some point as empty, for a previous task or otherwise. The car just handles that exception.

tatokis:
Thanks. I did some more research, and indeed it's used to be able to tell between empty strings and unset variables.
However, I see no reason to differentiate between the two, as both are invalid options in this scenario, and should be treated the same way.

For example `./quickinstall` should be treated the same way as `JOBS="" ./quickinstall`, which is what a simple `-z "$JOBS"` will do, since if $JOBS is unset, it will just end up being an empty string.

I don't see a reason to complicate this further.

Additionally, since this is a bash script, and I forgot to point it out last time, you should use [[ and ]] for consistency.

Thanks for the PR, and apologies for the extremely late response. Slowly trying to get things together.

PR #616, opened 2020-07-02
Fix #391 by replacing plain mutexes with FIFO-fair ones.

### LINKED ISSUE #391: Indicator LED activation delay
System: Gentoo Linux
Kernel: 4.19.27-r1 Gentoo Sources
Device: Corsair K68 Mechanical LED Keyboard
Driver Version: ckb-next 0.4.0
Build Type: Using the ./quickinstall script
Init system: Open RC

Current Behaviour:

When using the keyboard when toggling numlock, caps lock or scroll lock there is a delay of about 4/5 seconds between pressing the key and the relevant LED toggling on the keyboard to reflect the status.

Expected Behaviour:

When toggling the numlock, caps lock or scroll lock key the associated status LED should update immediately with the press of the relevant key.

skandalfo:
This fixes #391 by ensuring the RGB thread doesn't starve indicator updating by hoarding dmutex all the time.

The queued mutexes ensure FIFO locking order so the thread trying to set the indicator LEDs doesn't starve.

### DIFF
--- a/src/daemon/device.c
+++ b/src/daemon/device.c
@@ -8,14 +8,49 @@ int hwload_mode = 1;        ///< hwload_mode = 1 means read hardware once. shoul
 
 // Device list
 usbdevice keyboard[DEV_MAX];    ///< remember all usb devices. Needed for closeusb().
-pthread_mutex_t devlistmutex = PTHREAD_MUTEX_INITIALIZER;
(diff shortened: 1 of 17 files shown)


tatokis:
I rebased it (because I messed up my mutex branch) and merged it to master.
https://github.com/ckb-next/ckb-next/commit/ba063cedb751cecb1b8b1abf1af3f6e8dc31e9fe

PR #136, opened 2018-03-30
K95 Platinum HW Profile Support

ZirconiumX:
Very much work-in-progress

PR #627, opened 2020-08-03
Finish Dark Core support for the wireless branch

OrsanZ:
For now, this only fixes the lift height issue and removes the unnecessary DPI profiles.

What remains are:
- Animations in wireless
- DPI indicators 
- Battery status

Those are all related to the daemon, which is more complicated than I expected. I'm going to try to get at least two of the three done by Wednesday, but that is nothing more than an arbitrary deadline to keep me motivated.


src/daemon/dpi.c
```diff
@@ -175,6 +176,9 @@ int updatedpi(usbdevice* kb, int force){
         data_pkt[6] = (newdpi->x[i] >> 8) & 0xFF;
         data_pkt[7] = newdpi->y[i] & 0xFF;
         data_pkt[8] = (newdpi->y[i] >> 8) & 0xFF;
+        data_pkt[9] = light->r[LED_MOUSE + N_MOUSE_ZONES + i];
+        data_pkt[10] = light->g[LED_MOUSE + N_MOUSE_ZONES + i];
+        data_pkt[11] = light->b[LED_MOUSE + N_MOUSE_ZONES + i];
```
Ravenslofty, line 181: This modifies the packets for *all* mice, not just the Dark Core. In the spirit of trying not to break what presently works, could you wrap this in an `if (IS_DARK_CORE(kb))` please? I don't trust Corsair enough to assume that adding these packets won't break something.

Ravenslofty:
Additionally, I'm not too happy with the GUI changes, but it's down to @tatokis what to do there.

src/daemon/dpi.c
OrsanZ: Sure thing.

OrsanZ:
> In addition to the nit with changing the packets for all devices, you've also autoformatted usb.h, which adds noise in an unrelated pull request; please fix it.

I'm not too happy with this PR either. I was fighting with git and formatter, which made things really messy. The autoformatting was fixed in 94e598c498fb06f47996f425bcb6bd9acb94cb0c, but that only added to the noise. If you want we can close this PR and start from scratch.

Ravenslofty:
Try using `git rebase -i` to squash everything together; that might be a bit less noisy.

OrsanZ:
Alright, I finally cleaned up all the mess.

Ravenslofty:
Heads up: the merge of #628 will probably need a rebase or something.

OrsanZ:
Okay, I added the hwanim command, as @tatokis suggested. It really doesn't do anything fancy, it just takes a few parameters and passes them onto the device. I can try to do the GUI adjustments for hardware animations aswell.

Apart from that, this PR is pretty much done. A few small issues still exist, which are:
- The cursor freezes in place when the battery status is being fetched. This is pretty annoying after a while.
- First indicator light doesn't turn on when the mouse returns from sleep. We could use the Dark Core's native sleep function to prevent this.

Ravenslofty:
> The cursor freezes in place when the battery status is being fetched. This is pretty annoying after a while.

Maybe you could get a Wireshark trace; that might give some hints.

> First indicator light doesn't turn on when the mouse returns from sleep. We could use the Dark Core's native sleep function to prevent this.

Sleep on Linux is invisible to programs, so it's a tricky problem. The first time you know you went to sleep is after you wake up.

OrsanZ:
> Maybe you could get a Wireshark trace; that might give some hints.
I can try, but I think that this is normal behavior for usbrecv, as the same freezing occurs when starting the daemon.
> Sleep on Linux is invisible to programs, so it's a tricky problem. The first time you know you went to sleep is after you wake up.
I mean the setting where the lights go off after a certain amount of time, should've worded it better.

Ravenslofty:
> I think that this is normal behavior for usbrecv

No, this really isn't. Basically all the waiting usbrecv does is on the mouse; unless there's an unknown bug, all it does is send a message and wait for a reply. The Wireshark trace should tell us for sure if the mouse is blocking things.

OrsanZ:
> No, this really isn't.

I'll get on it then. :slightly_smiling_face:

OrsanZ:
Here you go.
[battery-issue.zip](https://github.com/ckb-next/ckb-next/files/5043655/battery-issue.zip)

tatokis:
The waiting happens in usbrecv. The reason it locks up like this is because the main input thread receives the reply, and then waits for the other thread to be ready to send the data to using the intmutex, but it's waiting instead. We should probably just remove all delay in usbrecv. If we're using waiting for the main thread. (same check as in usbsend that uses interrupts.) it doesn't matter as the input thread is the one receiving the data anyway. 

There's a DELAY_MEDIUMin usbrecv. 

This fix should go into master as it's needed for bragi as well.

OrsanZ:
Removing the DELAY_MEDIUM indeed fixes the issue.

OrsanZ:
That's fine. I believe that brightness is already supported on the K63WL then.

OrsanZ:
Okay, I'm done with the last thing, besides the hwanim GUI, I wanted to add.
@tatokis expressed his "interest" in doing the GUI. If he doesn't want to, I wouldn't mind doing it myself.

I'm also still having issues with CkbSettings::set in kbwidget.cpp on line 304. Saving the setting doesn't work, but that's the only real issue left. 

By the way, QTCreator shuffled some lines in kbwidget.ui, so there's a few unnecessary changed from that. If that's too much of a mess, I could go back and edit it manually.

internalc:
I was looking at the UI implementation and it come to my mind that perhaps a clean implementation of the hardware animation would be:

- daemon advertise **hwanim** feature in **dev/features**
- daemon create **dev/hwanim**, reading returns supported hw animations, one per line:
  `[animation code]:[animation name]`

Kb class would prepare the animations UI based on the hwanim feature. Alternatively the animation names could be hardcoded in the UI, however doing it this way seems more generic.

K63WL specific: The backlight intensity is controlled by a different command, however for UI compatibility we might also send the RGB values with **hwanim** cmd and translate it by some f(rgb)

OrsanZ:
Marking this as ready to review since it's pretty much done.

src/daemon/CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ if (MACOS OR LINUX)
               led_keyboard.c
               led_mouse.c
               led_mousepad.c
-              led_wireless.c
+              led_hw.c
```
Ravenslofty, line 80: Nitpicking a little bit, but the old name was better. Consider #136 which is *also* a hardware animation system but works very differently.

src/daemon/device_mouse.c
```diff
@@ -75,6 +75,25 @@ int setactive_mouse(usbdevice* kb, int active){
     return 0;
 }
 
+// FIXME: every time the battery is read, the mouse doesn't accept input, which makes using it quite uncomfortable to use
```
Ravenslofty, line 78: Has this been remedied or worked around? I can see us getting a *bunch* of issues about the mouse not working due to this.

src/daemon/device_mouse.c
```diff
@@ -75,6 +75,25 @@ int setactive_mouse(usbdevice* kb, int active){
...
+void* read_battery(void* context){
+    usbdevice* kb = context;
+    uchar msg[MSG_SIZE] = { CMD_GET, FIELD_BATTERY };
+
+    // Loop while device isn't being closed
+    while(kb->read_battery) {
+        uchar in[MSG_SIZE] = {};
+        if(!usbrecv(kb, msg, in)){
```
Ravenslofty, line 86: This is going to burn CPU cycles if `usbrecv` fails for some reason. Perhaps insert a `sleep(10)` here too?

Ravenslofty (changes requested):
I've got a few nits for the daemon I'd like looking at.

src/daemon/device_mouse.c
OrsanZ: Yes. This is the issue with DELAY_MEDIUM.

Ravenslofty: If so, could you remove the FIXME comment?

src/daemon/CMakeLists.txt
OrsanZ: Agreed.

src/daemon/device_mouse.c
OrsanZ: It's still there because the fix isn't committed yet. I'll add it and remove the comment.

src/daemon/device_mouse.c
OrsanZ: Fixed in new commit

OrsanZ:
> I'm also still having issues with CkbSettings::set in kbwidget.cpp on line 304. Saving the setting doesn't work, but that's the only real issue left.

By the way, this issue still exists, and I don't know how to fix it.

src/gui/kbwidget.cpp
```diff
@@ -281,6 +299,17 @@ void KbWidget::modesList_reordered(){
     currentProfile->modes(newModes);
 }
 
+void KbWidget::batteryTrayBox_stateChanged(int state){
+    QString batterySettingsPath("Devices/%1/batteryTrayIcon");
+    CkbSettings::set(batterySettingsPath.arg(device->usbSerial), state);
```
tatokis, line 304: The reason this "doesn't work", is because you're not supposed to access the `Devices` config group directly.

The device class overwrites everything every time it has to save something, and something it doesn't know about gets removed.

In order to store per-device settings, they need to be stored locally in the `Kb` class and then saved/loaded using the `Kb::save()` and `Kb::load()` functions, which are called automatically when the GUI autosaves.
The autosave takes place when a setter function in the class is called that sets _needSave to true.

You should save your setting at the end of the Kb::save function and restore it at the end of the load().
https://github.com/ckb-next/ckb-next/blob/e961dd02df7b97ce59c5c33cb65b8cc6b7219cb6/src/gui/kb.cpp#L311-L326

And finally access it here through the Kb class.

tatokis (changes requested):
This is what I have so far from a relatively quick glance.
I will have to rebase the wireless branch, as there have been some changes we'll need to take into account before merging this.

Additionally, regarding the removed DPI stages and lift height modifications, while the user can't directly modify them now that they are no longer part of the UI, a user can still import a profile that's aimed for another device, and that will add and possibly set those hidden options and confuse the user.

For example, the DPI stage arrow can end up pointing to a DPI stage at the bottom that doesn't exist.
I encountered this issue in the m45 branch but I have been unable to find an elegant solution.

Please have a look at my GUI changes for it, and if you can figure out a solution for it, open a PR against that branch, I'll merge it and then merge the m45 branch to master as that's been long pending.

The idea was to have the daemon expose how many DPI stages there are, and then have the GUI follow that, but from memory the value was hardcoded and const in multiple places.

tatokis:
Additionally, as @internalc said

> daemon advertise hwanim feature in dev/features
If this doesn't happen, it should. We should also have a way for the GUI to tell if we're plugged in via dongle or not.

> daemon create dev/hwanim, reading returns supported hw animations, one per line:
[animation code]:[animation name]

While this is an excellent idea, I'm not sure it's necessary. The idea of the daemon is to abstract features, so as long as both the gui and the know what the animation IDs are, it should be okay, especially since animation names need to be translatable.
That said, perhaps we can expose a list of animation ids/codes that are supported by this device so that the GUI can tell which ones to expose.

This isn't high priority right now though, and can be done later.

tatokis:
And one last thing. Once this is finished, please compile and test all the features this PR brings with both `-DSANITIZE_THREAD=1` and `-DSANITIZE_ADDRESS=1` individually, to make sure that we haven't introduced any new issues that have been missed.

It'd be a bit more difficult for me to do as I don't think I have FaKeyboard implemented enough to handle all of these cases.

OrsanZ:
I'll try to address the review this weekend, since I really don't have a lot of time right now.

> > daemon advertise hwanim feature in dev/features
>
> If this doesn't happen, it should. We should also have a way for the GUI to tell if we're plugged in via dongle or not.

There are already features in place for this. The feature "detachable" signals that the device can be detached or supports wireless and hwanims. The feature "wireless" means that a wireless dongle is connected. It's not really clear from the names, so if you can come up with better ones, I'll change them.

Ravenslofty:
The obvious thing then would be to rename "wireless" to "dongle" or
something.

src/daemon/led_wireless.c
```diff
@@ -1,62 +1,171 @@
...
+};
 
+enum HWANIM_SPEED {
+    HWANIM_SPEED_NONE   = 0,
+    HWANIM_SPEED_LOW    = 1,
+    HWANIM_SPEED_MEDIUM = 2,
+    HWANIM_SPEED_HIGH   = 3
+};
+
+int updatergb_wireless(usbdevice* kb, lighting* lastlight, lighting* newlight) {
```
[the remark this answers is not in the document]
OrsanZ: Maybe... Although the static color button uses the `rgb` command as well, so we'd have to change that to hwanim too.

src/daemon/led_wireless.c
```diff
@@ -1,62 +1,171 @@
...
 }
+
+void cmd_hwanim(usbdevice* kb, usbmode* dummy1, int dummy2, int keyindex, const char* code) {
+    // Unified format for the code string:
+    //  - DARK CORE Format: animation mode:speed:random:colour:colour2
+    //  - K63 WL    Format: animation mode:speed:dir
+    
+    uchar anim = 0xff, speed = 2, rand_or_dir = 0;
+    uchar r[2] = {}, g[2] = {}, b[2] = {};
+    sscanf(code, "%hhx:%hhu:%hhu:%02hhx%02hhx%02hhx:%02hhx%02hhx%02hhx", &anim, &speed, &rand_or_dir, r, g, b, r+1, g+1, b+1);
```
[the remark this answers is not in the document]
OrsanZ: While we certainly could do that, it would make it impossible to just leave out parameters you don't need for a certain animation mode. For example, `hwanim zone1:3:1` sets the animation to a rainbow at slow speed on zone 1. We don't need to pass a random/direction value or a color here. The variables are all initialized for this exact reason.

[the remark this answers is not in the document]
tatokis: Makes sense. As long as it doesn't send packets/change state by feeding it garbage, it should be okay.

src/daemon/led_wireless.c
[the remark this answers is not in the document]
tatokis: Yeah, that should be implemented using hwanim. The rgb command is used for spamming RGB software animations, and doing that through the dongle is a bad idea.
Some people use the daemon directly, and they most likely won't know they're not supposed to spam rgb data on these devices.

I was thinking we could allow SW animations when in wired, and HW animations when wireless. The new bragi devices do this as well, although differently.

src/daemon/devnode.c
```diff
@@ -191,6 +191,15 @@ static char* layoutstr(char layout){
     return "unknown";
 }
 
+void updatebattery(usbdevice* kb){
```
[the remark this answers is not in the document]
OrsanZ: Would using the `get` command for it be a good idea? Then we wouldn't need to use a thread for the daemon, which would fix the whole shutting down issue.

OrsanZ:
> Please have a look at my GUI changes for it, and if you can figure out a solution for it, open a PR against that branch, I'll merge it and then merge the m45 branch to master as that's been long pending.

That'd be the m45-test branch, correct?
I've tried to compile and got an error:
```
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/device_keyboard.c.o:(.rodata+0x0): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/device_mouse.c.o:(.rodata+0x18): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/devnode.c.o:(.data.rel.ro.local+0x0): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/firmware.c.o:(.rodata+0x1e0): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/main.c.o:(.rodata+0x8): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/notify.c.o:(.rodata+0x0): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/usb.c.o:(.rodata+0x40): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
/usr/bin/ld: CMakeFiles/ckb-next-daemon.dir/usb_linux.c.o:(.rodata+0x118): multiple definition of `devpath'; CMakeFiles/ckb-next-daemon.dir/command.c.o:(.rodata+0x120): first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [src/daemon/CMakeFiles/ckb-next-daemon.dir/build.make:450: bin/ckb-next-daemon] Error 1
make[1]: *** [CMakeFiles/Makefile2:375: src/daemon/CMakeFiles/ckb-next-daemon.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

OrsanZ:
Rebasing onto master solved it.

OrsanZ:
Okay, I changed the name for when the dongle is connected to `dongle`, as suggested. `wireless` is back to what it is in ckb-next:wireless. I made the suggested changes from the review as well.

src/gui/kb.cpp
```diff
@@ -323,6 +344,7 @@ void Kb::save(){
     settings.setValue("CurrentProfile", currentGuid);
     settings.setValue("Profiles", guids.trimmed());
     settings.setValue("hwLayout", KeyMap::getLayout(_layout));
+    settings.setValue("batteryIndicator", showBatteryIndicator ? "yes" : "no");
```
tatokis, line 347: Is there a reason you're not storing a bool? You can simply do 
`settings.setValue("batteryIndicator", showBatteryIndicator);`

src/gui/kb.cpp
```diff
@@ -283,6 +303,7 @@ void Kb::load(){
                 newCurrentProfile = profile;
         }
     }
+    showBatteryIndicator = QString::compare(settings.value("batteryIndicator").toString().trimmed().toLower(), "yes") == 0;
```
tatokis, line 306: This can be simplified if you just use a bool.
showBatteryIndicator = settings.value("batteryIndicator", true).toBool();

src/gui/ckbsystemtrayicon.cpp
```diff
@@ -25,7 +25,6 @@ void CkbSystemTrayIcon::setIcon(QIcon icon, QString name)
     }
 
     // If not, load the internal ones
-    qDebug() << "Indicator icon not found in theme hicolor";
```
tatokis, line 28: Why was this removed?

tatokis (changes requested):
One last thing. The icons should be rendered as pngs in appropriate sizes and installed into the system hicolor theme folders on linux, like it's being done for the ckb-next GUI icon.
Some DEs do not resize the tray icons themselves, and (especially with SVGs) you end up with a huge icon on your tray.
As an example, https://github.com/ckb-next/ckb-next/issues/630

Once these final concerns are addressed, I'll merge this.

src/daemon/devnode.c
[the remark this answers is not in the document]
tatokis: Late, but yeah, that's fine

src/gui/ckbsystemtrayicon.cpp
OrsanZ: Because it spammed the logs every time the icon updated. I'll add it again, since it should no longer print anything when the icon is installed into the theme.

tatokis: Ah, that makes sense. That thought didn't even cross my mind for some reason.

I did put it there so that it's easy to tell the fallback code is being used, as it forces a 22x22 resize, which will look pretty terrible on high dpi screens.
If after being readded it ends up being an issue, I'll find a way to make it only complain once, but for now, as you mentioned, it shouldn't be a problem.
Thanks!

OrsanZ:
Oh by the way, I ran my branch with the `-DSANITIZE_THREAD=1` and `-DSANITIZE_ADDRESS=1`.

`-DSANITIZE_THREAD=1` displays a bunch of data races, both in the GUI and the daemon, and lock-order-inversions in the GUI for me, both in master and in my wireless branch. Are these relevant?

`-DSANITIZE_ADDRESS=1` detects this leak when quitting the GUI on my wireless branch:

```
==13623==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 40 byte(s) in 1 object(s) allocated from:
    #0 0x7fd2dac47f41 in operator new(unsigned long) /build/gcc/src/gcc/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x560f63691c31 in Ui_KbWidget::setupUi(QWidget*) ckb-next/build/src/gui/ckb-next_autogen/include/ui_kbwidget.h:243
    #2 0x560f6368468a in KbWidget::KbWidget(QWidget*, Kb*) ckb-next/src/gui/kbwidget.cpp:23
    #3 0x560f63712fbd in MainWindow::addDevice(Kb*) ckb-next/src/gui/mainwindow.cpp:258
    #4 0x560f634a448a in MainWindow::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) ckb-next/build/src/gui/ckb-next_autogen/EWIEGA46WW/moc_mainwindow.cpp:164
    #5 0x7fd2d973006f  (/usr/lib/libQt5Core.so.5+0x2eb06f)
    #6 0x560f6349a762 in KbManager::kbConnected(Kb*) ckb-next/build/src/gui/ckb-next_autogen/EWIEGA46WW/moc_kbmanager.cpp:187
    #7 0x560f636394fa in KbManager::scanKeyboards() ckb-next/src/gui/kbmanager.cpp:193
    #8 0x7fd2d97257e1 in QObject::event(QEvent*) (/usr/lib/libQt5Core.so.5+0x2e07e1)
    #9 0x7fd2da4ad751 in QApplicationPrivate::notify_helper(QObject*, QEvent*) (/usr/lib/libQt5Widgets.so.5+0x15a751)

SUMMARY: AddressSanitizer: 40 byte(s) leaked in 1 allocation(s).
```

tatokis:
I didn't realise, but I forgot to specify to only run the daemon with those (and a GUI compiled without either of those).
There are a few false positives (and other unfixed issues in the GUI) that make it difficult to go through, as well as some wonky system libraries. Especially the GTK2 platform theme has a few issues that show up.

What you pointed out does seem like a valid memory leak, yeah, but probably not from your code, so it's unrelated to this PR.
If you want to attempt fixing it after this, you're welcome to do so. Otherwise I'll look into it later.
Off memory, there's also a race condition in the GUI relating to pulseaudio that I've also been meaning to look into.

There are two remaining data races in the daemon (in this branch) if I remember correctly, and both happen because of blocking threads. I did fix one of them in master, but there is still one left that's related to the device nodes in /dev. The one I fixed is in uinput. You can ignore those two.

Basically, as long as you don't get any issues specifically related to your code, it's all okay.

Thank you for taking the time!

OrsanZ:
How do I figure out if a data race is related to my code? For example, the battery icon is in the trace for one, but the last line says `SUMMARY: ThreadSanitizer: data race (/usr/lib/libQt5Core.so.5+0x2b625e)`. Does that mean that it's caused by my code?

tatokis:
That probably means that the function was called by the Qt main event loop. It doesn't mention the function because it can't find debug symbols for libQt5Core.so, but you probably don't need them.
Can you please post the whole trace?

OrsanZ:
```
WARNING: ThreadSanitizer: data race (pid=17670)
  Write of size 8 at 0x7b0800148480 by thread T19:
    #0 operator delete(void*, unsigned long) /build/gcc/src/gcc/libsanitizer/tsan/tsan_new_delete.cpp:150 (libtsan.so.0+0x8c85f)
    #1 <null> <null> (libQt5Gui.so.5+0x19a0fe)

  Previous write of size 8 at 0x7b0800148480 by main thread:
    #0 operator new(unsigned long) /build/gcc/src/gcc/libsanitizer/tsan/tsan_new_delete.cpp:64 (libtsan.so.0+0x8d1c4)
    #1 <null> <null> (libQt5Gui.so.5+0x19e627)
    #2 QtPrivate::FunctorCall<QtPrivate::IndexesList<0, 1>, QtPrivate::List<unsigned int, unsigned int>, void, void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int)>::call(void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int), BatteryStatusTrayIcon*, void**) /usr/include/qt/QtCore/qobjectdefs_impl.h:152 (ckb-next+0xed3ef)
    #3 void QtPrivate::FunctionPointer<void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int)>::call<QtPrivate::List<unsigned int, unsigned int>, void>(void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int), BatteryStatusTrayIcon*, void**) /usr/include/qt/QtCore/qobjectdefs_impl.h:185 (ckb-next+0xed3ef)
    #4 QtPrivate::QSlotObject<void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int), QtPrivate::List<unsigned int, unsigned int>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) /usr/include/qt/QtCore/qobjectdefs_impl.h:418 (ckb-next+0xed3ef)
    #5 <null> <null> (libQt5Core.so.5+0x2eb035)
    #6 Kb::readNotify(QString) ckb-next/src/gui/kb.cpp:557 (ckb-next+0xe2d60)
    #7 Kb::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) ckb-next/build/src/gui/ckb-next_autogen/EWIEGA46WW/moc_kb.cpp:142 (ckb-next+0x4738f)
    #8 QObject::event(QEvent*) <null> (libQt5Core.so.5+0x2e07e1)
    #9 __libc_start_main <null> (libc.so.6+0x28151)

  Thread T19 'Thread (pooled)' (tid=17691, running) created by main thread at:
    #0 pthread_create /build/gcc/src/gcc/libsanitizer/tsan/tsan_interceptors_posix.cpp:962 (libtsan.so.0+0x5efab)
    #1 QThread::start(QThread::Priority) <null> (libQt5Core.so.5+0xcd922)
    #2 MainWindow::MainWindow(QWidget*) ckb-next/src/gui/mainwindow.cpp:120 (ckb-next+0x16f347)
    #3 main ckb-next/src/gui/main.cpp:353 (ckb-next+0x41a2c)

SUMMARY: ThreadSanitizer: data race /build/gcc/src/gcc/libsanitizer/tsan/tsan_new_delete.cpp:150 in operator delete(void*, unsigned long)
```

src/gui/kb.cpp
OrsanZ: Interestingly enough, this is more equivalent to `!(QString::compare(settings.value("batteryIndicator").toString().trimmed().toLower(), "false") == 0);`

Because of that, keyboards would load with an enabled tray icon, which would lead to a segfault.
That would've happened before if I edited my conf file to enable the icon on a keyboard, so that'll be fixed too :sweat_smile:

OrsanZ:
Okay, I couldn't find one related to the battery with the aforementioned last line. Maybe I copied the wrong one.
However I found a data race that looks similar to the other one, but occurs when closing ckb-next.

```
WARNING: ThreadSanitizer: data race (pid=61054)
  Write of size 8 at 0x7b0c00155cd0 by thread T19:
    #0 operator delete(void*, unsigned long) /build/gcc/src/gcc/libsanitizer/tsan/tsan_new_delete.cpp:150 (libtsan.so.0+0x8c85f)
    #1 <null> <null> (libQt5Core.so.5+0xd1dd1)

  Previous write of size 8 at 0x7b0c00155cd0 by main thread:
    #0 operator new(unsigned long) /build/gcc/src/gcc/libsanitizer/tsan/tsan_new_delete.cpp:64 (libtsan.so.0+0x8d1c4)
    #1 QRunnable::create(std::function<void ()>) <null> (libQt5Core.so.5+0xcbb72)
    #2 KStatusNotifierItemPrivate::iconToVector(QIcon const&) ckb-next/src/gui/kstatusnotifier/kstatusnotifieritem.cpp:845 (ckb-next+0x1b9cc0)
    #3 KStatusNotifierItem::setToolTip(QIcon const&, QString const&, QString const&) ckb-next/src/gui/kstatusnotifier/kstatusnotifieritem.cpp:302 (ckb-next+0x1bacfc)
    #4 BatteryStatusTrayIcon::setBattery(unsigned int, unsigned int) ckb-next/src/gui/batterysystemtrayicon.cpp:16 (ckb-next+0x812e5)
    #5 QtPrivate::FunctorCall<QtPrivate::IndexesList<0, 1>, QtPrivate::List<unsigned int, unsigned int>, void, void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int)>::call(void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int), BatteryStatusTrayIcon*, void**) /usr/include/qt/QtCore/qobjectdefs_impl.h:152 (ckb-next+0xed3ef)
    #6 void QtPrivate::FunctionPointer<void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int)>::call<QtPrivate::List<unsigned int, unsigned int>, void>(void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int), BatteryStatusTrayIcon*, void**) /usr/include/qt/QtCore/qobjectdefs_impl.h:185 (ckb-next+0xed3ef)
    #7 QtPrivate::QSlotObject<void (BatteryStatusTrayIcon::*)(unsigned int, unsigned int), QtPrivate::List<unsigned int, unsigned int>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) /usr/include/qt/QtCore/qobjectdefs_impl.h:418 (ckb-next+0xed3ef)
    #8 <null> <null> (libQt5Core.so.5+0x2eb035)
    #9 Kb::readNotify(QString) ckb-next/src/gui/kb.cpp:557 (ckb-next+0xe2d60)
    #10 Kb::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) ckb-next/build/src/gui/ckb-next_autogen/EWIEGA46WW/moc_kb.cpp:142 (ckb-next+0x4738f)
    #11 QObject::event(QEvent*) <null> (libQt5Core.so.5+0x2e07e1)
    #12 __libc_start_main <null> (libc.so.6+0x28151)

  Thread T19 'Thread (pooled)' (tid=61083, running) created by main thread at:
    #0 pthread_create /build/gcc/src/gcc/libsanitizer/tsan/tsan_interceptors_posix.cpp:962 (libtsan.so.0+0x5efab)
    #1 QThread::start(QThread::Priority) <null> (libQt5Core.so.5+0xcd922)
    #2 MainWindow::MainWindow(QWidget*) ckb-next/src/gui/mainwindow.cpp:120 (ckb-next+0x16f347)
    #3 main ckb-next/src/gui/main.cpp:353 (ckb-next+0x41a2c)

SUMMARY: ThreadSanitizer: data race /build/gcc/src/gcc/libsanitizer/tsan/tsan_new_delete.cpp:150 in operator delete(void*, unsigned long)
```

tatokis:
Both of those seem "normal" to me. I think that's just Qt deleting the objects from another thread.
The issue on that is brought up inside kstatusnotifier, so it's definitely not in your code.
It's just that your code is creating a new tray icon, which uses kstatusnotifieritem.

OrsanZ:
Alright, after trying to fix a bug for an hour that ended up being resolved by a reboot, everything should work. I'm probably gonna open another PR for the DPI profile and lift height thing.

PR #661, opened 2020-11-13
Backend: Select device for mute indicator

penguineer:
This is the first half (backend part) for #561, adding the code to store a default mute device in the profile, show the mute state based on this device and mute the correct device when the button is pressed.

I submit this part as a separate PR so that it can be taken into account when making changes (kbperf.cpp seeing those more often).

The UI for selecting the device is an ongoing discussion in #561

### DIFF
--- a/src/gui/kbperf.cpp
+++ b/src/gui/kbperf.cpp
@@ -523,7 +523,7 @@ void KbPerf::applyIndicators(int modeIndex, const bool indicatorState[]){
             lightIndicator("lock", iColor[LOCK][1].rgba());
     }
     if(iEnable[MUTE]){
-        switch(getMuteState()){
+        switch(getMuteState(iMuteDev)){
         case MUTED:
             lightIndicator("mute", iColor[MUTE][0].rgba());
             break;
--- a/src/gui/kbperf.cpp
+++ b/src/gui/kbperf.cpp
@@ -53,6 +53,7 @@ KbPerf::KbPerf(KbMode* parent) :
         // Turn all other indicators on
         iEnable[i] = true;
     }
+    iMuteDev = SINK;
 }
 
 KbPerf::KbPerf(KbMode* parent, const KbPerf& other) :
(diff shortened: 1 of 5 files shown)


src/gui/kbperf.cpp
```diff
@@ -133,6 +136,10 @@ void KbPerf::load(CkbSettingsBase& settings){
         }
     }
     SGroup group(settings, "Performance");
+    {
```
tatokis, line 139: There's no need to make a new scope here

src/gui/media.h
```diff
@@ -15,12 +15,20 @@
 
 #endif
 
+// Mute device selection is supported
+EXTERN_C bool isMuteDeviceSupported();
+// Device class for mute state
+ENUM_C(muteDevice) {
+    SINK=0,
```
tatokis, line 22: No need to = 0, as enums always start from 0.

src/gui/media_linux.cpp
```diff
@@ -10,12 +10,14 @@
...
 
-static muteState lastKnown = UNKNOWN;
+static muteState lastKnownSink = UNKNOWN;
+static muteState lastKnownSource = UNKNOWN;
 
 #ifdef USE_LIBPULSE
 static pa_context* paContext = nullptr; //Context for communicating with Pulse Audio.
 static qint64 reconnectTime = 0; //Time (in MSecs since epoch) to attempt reconnect.
 static QString defaultSink; //Name of sink being checked if muted.
+static QString defaultSource; //Name of sink being checked if muted.
```
tatokis, line 20: Comment should read `Name of source`, not sink

tatokis (changes requested):
Minor nitpicks, but otherwise looks good to me.

src/gui/kbperf.cpp
penguineer: I followed the other examples and it hides the local `s_muteDev`, that is only there for readability.

tatokis: If you are referring to other cases such as https://github.com/ckb-next/ckb-next/blob/8348873a1ad6ab9c471b738c188887c648073e5a/src/gui/kbperf.cpp#L239-L251

then it's used to explicitly call the SGroup destructor to go one level up in the QSettings group hierarchy.
https://github.com/ckb-next/ckb-next/blob/2dbf7221c416a7648981c1312b8ccc3f1a5ab48e/src/gui/ckbsettings.h#L99

penguineer: I see. I removed the extra scope.

penguineer:
Thank you for the review, I have added the requested changes.

tatokis (approved):

Issue #711, opened 2021-04-29
Branch k95-xt numpad issue 


duonghao314:
When I install Ckb in branch `k95-xt` my numpad zone got an issue. 
Numpad 7 -> Nothing do 
Numpad 8 -> Numpad 7
Numpad 9 -> Numpad 8
Numpad 0 -> Numpad 9
Numpad .  ->  Numpad 0

Is this error can be corrected manually. I am currently configuring with macros, but I cannot handle numpad 7 key because it not response any key.

TheWebMachine:
I, too, have this issue. Just compiled from the k95-xt branch today. Here's my diddly-do:
[ckb-next-report-1620015354.txt](https://github.com/ckb-next/ckb-next/files/6412823/ckb-next-report-1620015354.txt)
(NOTE: The service crash is related to trying to change the polling rate in the GUI, for which I will open a separate bug, but this numpad issue also plagues me.)

TheWebMachine:
Found another mis-mapped key. The Menu modifier key is mapping as `<` Unbinding does nothing as it isn't the right "scancode" when you press the key

TheWebMachine:
So, it looks like the only key left to sort out is numpad 7. Everything else is working great as of 12f582d

TheWebMachine:
Has any progress been made on numpad 7 and the Menu modifier? (I misspoke in my last comment that only numpad 7 was still broken; Menu modi is still coding as `<` and can't be re-mapped.) Anything I can do to help? 

I've taken to scripting a stop of the ckb service automatically when my IDEs are launched so I can write without losing my sanity over that stupid 7 key...never realized just how much I rely on numpad when I code. 🤣️ So, yeah, if there's anything I can do to help, I'm all ears!

tatokis:
I've been reworking a big chunk of the usb code in order to merge the k95-xt branch to master under the experimental device flag, as they had diverged a lot, and the hacks were piling up. I'll get to this after I've finished merging and cleaning up.

FryDay:
Any update on this @tatokis? No 7 key gets annoying after a bit.

tatokis:
@FryDay not really, sorry. I did receive a K57 which behaves the same way in wired mode, so once I get some basic support for it, I'll fix the numpad 7 issue.

tatokis:
https://github.com/ckb-next/ckb-next/issues/543#issuecomment-931025424
Please test again after reading the above.

Issue #543, opened 2020-02-07
K95 RGB Platinum XT / New Protocol


rporter220:
I just purchased a K95 RGB Platinum XT keyboard hoping that it would be compatible with ckb-next since the K95 Platinum is on the supported keyboard list. The keyboard does not seem to be picking up the keyboard with the latest git version of ckb-next.

I'm thinking that since this is technically a different model it probably has a different device ID or something along those lines.

If there is anything I can provide, please let me know and I assist any way I can to help getting this keyboard supported. :-)

Thank you!

rporter220:
If it is just a USB Device ID issue, I hope this helps:
USB Device ID reported is `1b1c:1b89`

tatokis:
The keyboard uses the completely new protocol that no one has implemented. Unless someone steps up and does it, it won't be supported for quite a while

rporter220:
Unfortunately I'm not a software developer, but is there any way I can assist for someone to start putting this in their pipeline?  

I'm willing to purchase a keyboard for someone to begin this or if there is anything else I can contribute, just let me know! :-)

rporter220:
file:///home/richard/Documents/ckb-next-report-1589319382.gz

rporter220:
[ckb-next-report-1589319382.txt](https://github.com/ckb-next/ckb-next/files/4618580/ckb-next-report-1589319382.txt)

tatokis:
Can you please dump the HID descriptors?
`sudo usbhid-dump -d 1b1c:1b89`

ghost1347:
Hey I noticed he has not gotten back to you yet, Can I post mine?

tatokis:
Sure

ghost1347:
003:006:003:DESCRIPTOR         1601607982.430057
 05 01 09 02 A1 01 85 05 09 01 A1 00 05 09 19 01
 29 20 15 00 25 01 95 20 75 01 81 02 05 01 09 30
 09 31 16 01 00 26 FF 7F 75 10 95 02 81 06 09 38
 15 81 25 7F 75 08 95 01 81 06 C0 C0

003:006:002:DESCRIPTOR         1601607982.433051
 06 42 FF 09 02 A1 01 15 00 26 FF 00 95 40 75 08
 09 02 81 02 C0

003:006:001:DESCRIPTOR         1601607982.436041
 06 42 FF 09 01 A1 01 15 00 26 FF 00 95 40 75 08
[...]

PR #715, opened 2021-05-03
Fix keymap for numpad 0.89

dcat52:
- Number pad 7 still not working


dcat52:
Reference to #543 and #711

PR #725, opened 2021-06-18
Fix slow power off and reboot for strafe keyboard.

TOKYEO:
I fixed a bug where power off and reboot was slow because the strafe keyboard will hang.

### DIFF
--- a/src/daemon/usb.h
+++ b/src/daemon/usb.h
@@ -201,7 +201,7 @@ const char* product_str(ushort product);
 
 /// Some devices cause usbhid to spend a long time initialising it. To work around this, we intentionally uncleanly
 /// deinitialise the device, skipping the usbhid handover.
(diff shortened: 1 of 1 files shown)


tatokis:
I don't remember this being an issue with the strafe. Have you tried the HID quirks first? If the quirks fix it, then the issue should be fixed in the kernel. I'd like to have as few devices as possible in the unclean exit list.
https://github.com/ckb-next/ckb-next/wiki/Troubleshooting#usbhid-quirks

TOKYEO:
Oh ok, thank you! I will try it out and let you know.

TOKYEO:
Really sorry to bother you, but what shall I do once I have found my specific USBCore/USBHID quirks?

TOKYEO:
Thank you for the input. Adding the HID to the kernel command line parameter worked. My reboot is now at a normal speed.

Thank you for your time,
TOKYEO

PR #728, opened 2021-06-29
Initial support for the Corsair Dark Core RGB Pro SE

olokos:
In order to test it you need to run the daemon with experimental flag
ex.
sudo ./bin/ckb-next-daemon --enable-experimental

Forwards and back button work, that's an improvement over no ckb at all.
RGB doesn't work
DPI indicator is stuck at 2 of 3 dots
DPI switching and DPI changes + Profiles seem to work as normal


olokos:
seems like scroll up doesnt work with CKB GUI running (software mode)
but with CKB GUI OFF, but daemon running with --enable-experimental (hardware mode)
scroll up/down works fine

helpful data:
Single event up and down
https://hastebin.com/pavojidota.yaml

Multiple events up and down:
(Might contain useless data because of micromovements)
https://hastebin.com/emetadiloq.yaml

PR #748, opened 2021-09-21
Add 66 service

ndowens:
This adds 66 script for ckb-next. I am using it and it works fine.

### DIFF
--- a/linux/66/ckb-next
+++ b/linux/66/ckb-next
@@ -1,0 +1,8 @@
+[main]
+@type = classic
+@description = "ckb-next daemon"
+@version = 0.0.1
+@user = ( root )
+
+[start]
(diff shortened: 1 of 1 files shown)


tatokis:
While I have never heard of it before, and not wanting to set up a test VM, I'll merge this despite not testing it.
You might want to look into installing it automatically with cmake, in a similar way to other init scripts, otherwise people might not know it exists.

ndowens:
> While I have never heard of it before, and not wanting to set up a test VM, I'll merge this despite not testing it.
> You might want to look into installing it automatically with cmake, in a similar way to other init scripts, otherwise people might not know it exists.

That is my next plan to do, after I was going somewhere I thought about it and decided I should have in the first place :) But I will open a PR once that is done.

PR #776, opened 2021-12-30
Use directly Q_SLOTS/Q_SIGNALS/Q_EMIT

Montel:


Ravenslofty:
Uh, why?

Montel:
Because it will be necessary for using "QT_NO_KEYWORDS" which will help to port Q_FOREACH and co for porting against qt6.

tatokis:
I don't see myself merging this. It feels like a headache to use the Q_ symbols every time.
As for porting to Qt6, can you not simply build against Qt6 and replace the removed macros like `foreach` as you go, wherever it complains?
If I recall correctly, there are also defines that you can use to prevent use of deprecated functions and macros without disabling every keyword.

Montel:
Ok as you want I will close this MR

Issue #440, opened 2019-07-03
Device Support Request - M55 RGB PRO


Unaccounted4:
Report:
[ckb-next-report-1562169446.gz](https://github.com/ckb-next/ckb-next/files/3355768/ckb-next-report-1562169446.gz)

Corsair M55 RGB PRO 

Keep on rocking!

tatokis:
Unfortunately, the bad news is that despite this being a wired mouse, Corsair are using the new protocol that we don't have support for yet. It will probably be a while before we get it working. I think they'll be switching most new devices to it going forward, so we'll eventually have to support it.

That said, can you please run `sudo usbhid-dump -d 1b1c:1b70`

And after that, compile this https://github.com/ckb-next/getUsbString and run

```
sudo ./getUsbString 0x1b1c 0x1b70 4
```
and
```
sudo ./getUsbString 0x1b1c 0x1b70 5
```

Please paste the output of all of those here.

Unaccounted4:
Here you go: 
`sudo usbhid-dump -d 1b1c:1b70
003:012:003:DESCRIPTOR         1562186967.142327
 05 01 09 06 A1 01 05 07 19 E0 29 E7 15 00 25 01
 75 01 95 08 81 02 19 00 29 97 15 00 25 01 75 01
 95 98 81 02 05 08 19 01 29 03 25 01 75 01 95 03
 91 02 75 05 95 01 91 01 C0

003:012:002:DESCRIPTOR         1562186967.142570
 06 42 FF 09 02 A1 01 15 00 26 FF 00 75 08 95 40
 09 02 81 02 C0

003:012:001:DESCRIPTOR         1562186967.142859
 06 42 FF 09 01 A1 01 15 00 26 FF 00 75 08 95 40
 09 01 81 02 95 40 09 01 91 02 C0

003:012:000:DESCRIPTOR         1562186967.143125
 05 01 09 02 A1 01 09 01 A1 00 05 09 19 01 29 20
 15 00 25 01 95 20 75 01 81 02 75 10 95 02 05 01
 09 30 09 31 16 01 80 26 FF 7F 81 06 75 08 95 01
 05 01 09 38 15 81 25 7F 81 06 C0 C0`

`sudo ./getUsbString 0x1b1c 0x1b70 4
Read 12 bytes
Raw (hex): 0c 03 56 00 30 00 33 00 30 00 30 00 
String: V0300`

`sudo ./getUsbString 0x1b1c 0x1b70 5
[...]

PR #778, opened 2022-01-03
Add Corsair M55 mouse support 

Halamix2:
[Mouse manual](https://www.corsair.com/corsairmedia/sys_master/productcontent/49_001821_revAA_NA_M55_RGB_PRO_QSG.pdf)
- [x] add daemon support
    - [x] add proper button LUT
    - [x]  add proper button/LED mapping
- [x]  add gui support
    - [x] add .svg file
- [x] default DPI action to "cycle up" in new modes
- [x] fix DPI LED update
    - kinda fixed, flashed slightly for 0.06-0.08s, I have no idea how to fix that, so it'll stay like this until someone who isn't me fixes that

**Linked issue:**
#440


src/daemon/keymap.c
```diff
@@ -939,6 +939,26 @@ const unsigned char harpoon_wl_lut[BRAGI_ONE_BYTE_MOUSE_BUTTONS] = {
...
+const unsigned char m55_wl_lut[BRAGI_ONE_BYTE_MOUSE_BUTTONS] = {
+    0x00,
+    0x01,
+    0x02,
+    0x04,
+    0x03,
+    0x06, //forwards
+    0x05, //backwards
+    0x08, //dpi up?
+    0x09,
```
tatokis, line 952: BRAGI_ONE_BYTE_MOUSE_BUTTONS is 8 but there are more than 8 entries here. The compiler should be throwing a warning.

src/daemon/keymap.c
```diff
@@ -948,13 +968,15 @@ void corsair_bragi_mousecopy(usbdevice* kb, usbinput* input, const unsigned char
...
     // We need a better way to identify this
-    if(kb->vendor == V_CORSAIR && (kb->product == P_DARK_CORE_RGB_PRO_SE || kb->product == P_DARK_CORE_RGB_PRO_SE_WL || kb->product == P_HARPOON_WL_U))
+    if(kb->vendor == V_CORSAIR && (kb->product == P_M55_RGB_PRO || kb->product == P_DARK_CORE_RGB_PRO_SE || kb->product == P_DARK_CORE_RGB_PRO_SE_WL || kb->product == P_HARPOON_WL_U))
         buttons = BRAGI_ONE_BYTE_MOUSE_BUTTONS;
 
     // Pick the appropriate LUT. We can't patch the keymap as that will break standard HID input.
     const unsigned char* lut = corsair_bragi_lut;
     if(kb->vendor == V_CORSAIR && kb->product == P_HARPOON_WL_U)
         lut = harpoon_wl_lut;
+    if(kb->vendor == V_CORSAIR && kb->product == P_M55_RGB_PRO)
```
tatokis, line 978: This should be an 'else if'

src/daemon/keymap_patch.c
```diff
@@ -16,6 +16,14 @@ keypatch k63patch[] = {
 };
 #define K63PATCH_LEN sizeof(k63patch)/sizeof(*k63patch)
 
+keypatch m55patch[] = {
+    { 210, "mouse6", -1, KEY_CORSAIR },
+    { 211, "mouse7", -1, KEY_CORSAIR },
+    { 213, "dpiup",  -1, KEY_CORSAIR },
+    { 237, "dpi", LED_MOUSE, KEY_NONE },
```
tatokis, line 23: Please at least zero out the existing dpi entry as well, because as it is, there are two "dpi" leds with this patch applied. This might also be the reason the DPI led wasn't updating correctly.

src/gui/keymap.cpp
```diff
@@ -375,6 +375,25 @@ static const Key M95Keys[] = {
 };
 #define KEYCOUNT_M95    (sizeof(M95Keys) / sizeof(Key))
 
+
+// M55
+// TODO
```
tatokis, line 380: I don't think this is a todo anymore.

tatokis (changes requested):

src/gui/keymap.cpp
Halamix2: removed TODO line

src/daemon/keymap_patch.c
Halamix2: Shortened LUT to 8 entries

src/daemon/keymap.c
Halamix2: Shortened LUT to 8 entries

src/daemon/keymap.c
Halamix2: Added else if

wimvelzeboer:
@Halamix2 @tatokis Could you please resolve the merge conflict, that it can be merged? I have a M55 and would really like to use it properly in Unbuntu. Thanks!

noneistaken:
Works well, except that the DPI toggle button doesn't cycle the last mode back to the first mode, it's just stuck here
image

tatokis:
You can rebind it to DPI Cycle manually for now in the binding tab.
Other than that, if you don't have any animations set, does the DPI LED update accordingly? Or do you have to change the brightness for it to take effect?

noneistaken:
I do have to change the brightness or have some animation playing, else it will flash to the assigned colour and back to white whenever I cycle the DPI.

noneistaken:
Why won't this be merged

Halamix2:
For me the DPI cycles back to first position.
My mouse firmware: 4.07

wimvelzeboer:
> 

Cool, thanks for the tip! Now everything works for me perfectly! 
For me the DPI LED does work as expected.

Halamix2:
DPI LED updates now, but it flashes for 0.6-0.8s before updating the color. I have no idea how to fix that, so feel free to create  a RP with the fix :)

noneistaken:
I guess you can try saving the DPI profile to the mouse directly, with the colours and all.

PR #795, opened 2022-02-23
WIP: K60 pro rgb

akatran:
2 Issues

1. Modifier keys does not light up,
2. ckb-next gui : Could not display other layouts except English (United states) & English (United states, Dvorak)


akatran:
That does not seem right!

PR #796, opened 2022-02-23
Add K60 PRO RGB support

akatran:
2 Issues

1. Modifier keys does not light up,
2. ckb-next gui : Could not display other layouts except English (United states) & English (United states, Dvorak)


akintelligentmedia:
I was capturing packages with wireshark and when I try only the problematic keys I saw that there where no messages passing from GUI to daemon!

Is it something else that needs to be mapped in order to make these keys working?

akintelligentmedia:
Also my K60 is like this..
K60PRORGB
.

akatran:
I was capturing packages with wireshark and when I try only the problematic keys I saw that there where no messages passing from GUI to daemon!

Is it something else that needs to be mapped in order to make these keys working?

akatran:
Also my K60 is like this.
K60PRORGB
.

tatokis (changes requested):
Looks pretty good, but there are a few last things that need to be looked at.

You should probably add the K60 in DEV_HAS_VOLWHEEL() (there's a not there, so any device listed does _not_ have a volume wheel).
You should also check if it supports adjusting the pollrate through software. You can add it to SUPPORTS_ADJRATE() and then check if you can actually change the pollrate through the GUI (dropdown in device tab).

Finally, it'd be appreciated if you could either squash these commits, or combine them in some way and then force push to your branch. It will make things easier in the future for someone potentially bisecting the code.

akatran:
Squash went totally wrong. 
I'll fix it first and re-apply the PR again

akintelligentmedia:
I notice the same but I was unsure if it was due to switching to the Zen kernel.
Now that you also have the same issue and hopefully you didn't switch to the Zen Kernel like I did I'll give it a shot why it fails to detect the keyboard!

Thanks for pointing this out.

WhateverFits:
OK, so my issue was kinda stupid. Somehow the brightness level got turned to Off. I just now saw that after a few weeks of it not working. I changed it to 33% and I've got a lit up keyboard again.

WhateverFits:
I would think it was USB permission related except that it manages to turn
off the LEDs on the keyboard when it engages. The LEDs turn back to their
set color when switching over to a different computer or when ckb isn't
running yet on boot.

PR #656, opened 2020-11-08
Nightsword RGB support

openaiken:
Hello! I have added Nightsword RGB support. All critical functionality has been added (all lighting zones, and binding for all[*] keys).

This does not add support for some of the model specific features seen in iCUE, such as the visual mapping of the consumer-configurable counter-weights (accessed from a panel on the bottom of the device), or the calibration feature (where you drag the cursor in a spiral to adjust x/y sensitivity after adjusting counter-weights). This is due to both A) these features not being supported by the version of the corsair protocol that ckb-next uses, and B) not having these features designed in the GUI (and hooboy do I NOT want to be the one to do that!).

[*] The key that there is not binding for yet is the Profile Up/Down switches. The Profile Up switch is detected as "profswitch" (see /src/daemon/keymap.c) as verified by monitoring /dev/input/ckbX/notify1, and the Profile Down switch is not detected at all. This is probably not a terribly complicated fix, but it involves touching stuff related to the daemon and the protocol, which I'm not familiar with yet. On my personal profile for the mouse I just set the Profile Up key to cycle through profiles.

PR #798, opened 2022-02-26
Add K60 PRO RGB support

akatran:
Now it has all the changes and squashed everything to one commit!


akatran:
I've notice that the multimedia keys does not work while app is running.
Is there a special place that need to be declared these?

tatokis:
There are two possibilities.
All keyboards up to now without dedicated media keys have been sending media key scancodes when you'd press the Fn + FX combinations. It is possible your keyboard does this and it's not implemented correctly. Are there any errors/warnings in the daemon? You'd need to use wireshark and look at the input packets when you press the keys to see what happens and what changes.

When in software mode (GUI running), you get two packets for each key action. So for a keydown and a keyup, that's 4 packets in total. You need to look at the changes in the long input packet (that's the software one).

The other possibility is that it expects the software to do the action on the fly, however since ckb does not support macros with key combinations, we can't really do much unless someone implements it.

akatran:
It is a bit different on software mode than when running without.
This is the packages that are send on both situations.
image


It seems that without ckb-next it just sends the scan codes and that is displayed above. The 4 entries are 2 for pressing just the F8 key and another two for FN+F8. On both of the cases the first is the key-down and the second is the key-up as you said.
No 13 payload : 010000000000000000000200000000000000000000
No 13 payload : 
No 13 payload : 



image

On the other hand you can see the messages while the ckb-next is running and the messages are : 
No 13 + 14  : F8 keydown
No 17 + 18 : F8 keyup
No 21 : FN keydown
No 23 : F8 keydown
No 25 : F8 keyup
No 27 : FN keyup

akatran:
Also there are no errors displayed on the daemon.

tatokis:
Unfortunately, it is as I suspected. The media keys are implemented in software.

Since Fn and F8 respectively are:
```
00020000000000000000000000000000000400000000000000000000000000...
00020000000000000000020000000000000000000000000000000000000000...
```
When you combine them, you get this, which just tells the software that "Fn and F8 are both pressed at the same time":
```
00020000000000000000020000000000000400000000000000000000000000...
```

Whereas a scancode for volume up should instead be:
```
00020000000000000000000000008000000000000000000000000000000000...
```

So the fact that you are not getting the scancode for volume up means that it's software doing multi-key macros, which ckb doesn't support. Does your keyboard, without any software installed (ckb-next-daemon stopped) still have functional volume keys?
If so, we could make a hack in the daemon for each media key that is along the lines of: "If Fn + F8, then pretend neither of them are pressed down, and instead pretend Volume Up is pressed", but that means that the user can't rebind them at all, and in general is a really ugly workaround.

This is really unfortunate, but I also noticed that there are no media key markings on your F keys like there are on a keyboard like the strafe [0], so maybe they are intentionally doing it this way.

[0] https://b.scdn.gr/images/sku_main_images/008685/8685072/20160330120700_corsair_strafe_rgb_mechanical.jpeg

tatokis:
> The USB HID detects that someone pressed FN + Fx key and change the response to the appropriate keycode?

Yes, that is what is done through iCUE.
If you don't run ckb-next, the keyboard runs in hardware mode, and instead sends a scancode that means volume up in the first place.

I have pushed a commit in the branch. It should pick up the media keys from the "hardware" packets as a workaround.
It won't support unbinding, but it's more important to have media keys for now than not being able to unbind them.

Can you please test this?

akatran:
I'm fine with having the media keys and miss the re-binding feature.
Unfortunately it does not work!

It seams like the Fn key does not count as pressed!

Also while in "software" mode I notice that the Fn key is send separately from the other key, F7 and F8 in our case for volume down and volume up.

src/daemon/keymap.c
```diff
@@ -682,7 +682,7 @@ void process_input_urb(void* context, unsigned char* buffer, int urblen, ushort
                     if(urblen == 64) {
                         corsair_kbcopy(targetkb->input.keys, buffer + 2);
                     } else if(firstbyte == NKRO_KEY_IN || firstbyte == NKRO_MEDIA_IN) {
-                        if(!targetkb->active)
+                        if(!targetkb->active || (firstbyte == NKRO_MEDIA_IN && BRAGI_HAS_MEDIA_MACRO(targetkb)))
```
akatran, line 685: In my case at least this is not going to be called as the Wireshark displays "Bragi extra key input" 
and the "Extra HID" is as below:
```
0000   00 2c c9 85 b0 99 ff ff 43 01 83 04 08 00 2d 00   .,......C.....-.
0010   7c 19 29 62 00 00 00 00 3c f9 08 00 00 00 00 00   |.)b....<.......
0020   40 00 00 00 40 00 00 00 00 00 00 00 00 00 00 00   @...@...........
0030   01 00 00 00 00 00 00 00 00 02 00 00 00 00 00 00   ................
0040   00 02 00 00 00 00 00 00 00 00 02 00 00 00 00 00   ................
0050   00 04 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0060   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0070   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................

```

That will satisfy both the `kb->protocol == PROTO_BRAGI` and the `urblen == 64` so it will call the `corsair_kbcopy` function.

tatokis: urblen is not 64 in hardware media key packets. If I am not mistaken, they are either 3 or 4 bytes.
Please try capturing volume up without ckb-next running.

I am starting to suspect though that for some reason hardware mode media key packets aren't sent in software mode, which would explain why this doesn't work. For clarification the extra key input packets are the software packets, and the HID input ones are the hardware ones.

akatran:
You are right!
```
0000   c0 00 9b 83 af 99 ff ff 43 01 81 04 08 00 2d 00
0010   8d 20 29 62 00 00 00 00 df 6a 06 00 00 00 00 00
0020   05 00 00 00 05 00 00 00 00 00 00 00 00 00 00 00
0030   01 00 00 00 00 00 00 00 04 02 00 00 00 00 00 00
0040   02 e9 00 00 00
```

akatran:
The above was with ckb-next-daemon disabled!

egberts:
HOWTO get keyscan code


https://github.com/ckb-next/ckb-next/pull/656#issuecomment-729265011

PR #800, opened 2022-03-08
Fix maybe-uninitialized warning

ericherman:
Initialize to zero at declaration to remove warnings:

```
/home/eric/src/ckb-next/src/gui/kb.cpp: In member function ‘void Kb::readNotify(const QString&)’:
/home/eric/src/ckb-next/src/gui/kb.cpp:742:24: warning: ‘y’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  742 |                 int x, y;
      |                        ^
/home/eric/src/ckb-next/src/gui/kb.cpp:742:21: warning: ‘x’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  742 |                 int x, y;
      |                     ^
```

### DIFF
--- a/src/gui/kb.cpp
+++ b/src/gui/kb.cpp
@@ -739,7 +739,7 @@ void Kb::readNotify(const QString& line){
                 if(dpi.length() != 2)
                     continue;
                 QStringList xy = dpi[1].split(',');
-                int x, y;
+                int x = 0, y = 0;
                 bool off = false;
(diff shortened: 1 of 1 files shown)


tatokis:
I'll merge this purely to silence the warning.
The variables are not used uninitialised, however the compiler can't really tell, thus the *may* in the warning.
This code should be refactored at some point, but I don't really have the time or energy to deal with it.

Issue #74, opened 2018-02-05
K63 Wireless support


felbus:
Hi,

I am getting no devices connected when connecting this device. I have read all other issues, and it seems support was added, but I am having no luck.

ckb-dev-detect-report is at bottom of this issue.

I have tried building from quickinstall and using the package. Have the latest xcode, and have plugged in to windows and upgraded firmware.

Have removed ckb repo, and qt5, then reinstalled, rebooted, and tried all other suggestions, but still getting No devices connected.  

Current behaviour is that keys work, and are mapped to mac layout (keyboard is US layout), led is working. I can plug in to a windows machine, change the lighting, then put it back in mac and it is saved. 

ckb-dev-detect-report:
```ckb-dev-detect - Mon Feb  5 19:34:11 GMT 2018

Kernel: Darwin X-MacBook-Pro.local 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64

OS Version:
ProductName:	Mac OS X
ProductVersion:	10.13.3
BuildVersion:	17D47

USB Devices:
                CORSAIR K63 Wireless Mechanical Gaming Keyboard:

                  Product ID: 0x1b45
                  Vendor ID: 0x1b1c  (CORSAIR MEMORY INC.)
                  Version: 1.02
                  Serial Number: 1401A00BAF15842759618230F5001BC4
                  Speed: Up to 12 Mb/sec
                  Manufacturer: CORSAIR
                  Location ID: 0x14142000 / 10
                  Current Available (mA): 500
                  Current Required (mA): 500
                  Extra Operating Current (mA): 0
```

ZirconiumX:
The daemon doesn't have the USB product ID yet. That in itself isn't a major problem. But they might have tried to change something, so you should join us in IRC.

tatokis:
@felbus Can you please try the code from the PR above and check if it works?

felbus:
Hi, it was plugged in.

ZirconiumX:
Can you take another ckb-dev-detect with it in wireless mode?

felbus:
Yep, ill have to do tonight when I get home. Thanks for looking.

ZirconiumX:
@felbus Bump.

felbus:
Wireless https://pastebin.com/1iBnMPqB

felbus:
[...]

PR #817, opened 2022-04-16
K63 wireless layout should be the same as K65

Grawp:
K63 wireless layout should be the same as K65 including the existence of a JP variant.

Discussed in https://github.com/ckb-next/ckb-next/issues/74. Only the wired K63 does not have the FN key.

### DIFF
--- a/src/gui/keymap.h
+++ b/src/gui/keymap.h
@@ -83,6 +83,7 @@ public:
         M55,
         K60,
         K57_WL,
+        K63_WL,
         _MODEL_MAX
     };
     // Key layouts (ordered alphabetically by name)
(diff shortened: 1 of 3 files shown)


src/gui/keymap.h
```diff
@@ -54,6 +54,7 @@ class KeyMap {
         NO_MODEL = -1,
         K55,
         K63,
+        K63_WL,
```
tatokis, line 57: https://github.com/ckb-next/ckb-next/blob/e2d81007067df6de11d9b0eb99a72f928551b53d/src/gui/keymap.h#L51-L52

Please add it right before _MODEL_MAX.

tatokis (changes requested):

src/gui/keymap.h
Grawp: Done.

Issue #834, opened 2022-05-28
Broken install cannot be removed


docsmooth:
**Click the `Generate Report` button in the `Settings` tab of the ckb-next user interface and attach the generated report to this issue.**

## General information
ckb-next 0.5.0.1~impish
Ubuntu 20.10 (yes, I know)

## Logs & crash reports
$ sudo dpkg --configure -a
Setting up ckb-next (0.5.0-1~impish) ...
Failed to enable unit: Unit file ckb-next-daemon.service does not exist.
Failed to start ckb-next-daemon.service: Unit ckb-next-daemon.service not found.
dpkg: error processing package ckb-next (--configure):
 installed ckb-next package post-installation script subprocess returned error exit status 5
Processing triggers for desktop-file-utils (0.26-1ubuntu2) ...
Processing triggers for hicolor-icon-theme (0.17-2) ...
Processing triggers for gnome-menus (3.36.0-1ubuntu1) ...
Processing triggers for mailcap (3.69ubuntu1) ...
Processing triggers for bamfdaemon (0.5.5+21.10.20210710-0ubuntu1) ...
Rebuilding /usr/share/applications/bamf-2.index...
Errors were encountered while processing:
 ckb-next
$ sudo dpkg-reconfigure ckb-next
/sbin/dpkg-reconfigure: ckb-next is broken or not fully installed
$ sudo apt -y remove ckb-next
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following packages will be REMOVED:
  ckb-next
0 upgraded, 0 newly installed, 1 to remove and 31 not upgraded.
1 not fully installed or removed.
After this operation, 3137 kB disk space will be freed.
(Reading database ... 528581 files and directories currently installed.)
Removing ckb-next (0.5.0-1~impish) ...
Failed to stop ckb-next-daemon.service: Unit ckb-next-daemon.service not loaded.
Failed to disable unit: Unit file ckb-next-daemon.service does not exist.
dpkg: error processing package ckb-next (--remove):
 installed ckb-next package pre-removal script subprocess returned error exit status 1
dpkg: too many errors, stopping
Failed to start ckb-next-daemon.service: Unit ckb-next-daemon.service not found.
dpkg: error while cleaning up:
 installed ckb-next package post-installation script subprocess returned error exit status 5
Errors were encountered while processing:
 ckb-next
Processing was halted because there were too many errors.
E: Sub-process /usr/bin/dpkg returned an error code (1)

[...]

PR #835, opened 2022-05-28
fix for #834 postinstall fails if install is broken

docsmooth:
1) removing sudo from inside the postinstall - dpkg or apt should already be running as root, and some system policies don't allow root to call sudo (causes audit tracability issues)

2) add some if statements to change the exit status if ckb-next binary doesn't exist, so that removal can be clean

### DIFF
--- a/src/animations/mviz/main.c
+++ b/src/animations/mviz/main.c
@@ -283,7 +283,7 @@ static bool mviz_init_capture(){
     }
 
     if(!pa_str){
-        static_assert(sizeof(float) == 4);
+        static_assert(sizeof(float) == 4, "Expected 32 bit floats");
         static const pa_sample_spec ss = {
             .format = PA_SAMPLE_FLOAT32LE,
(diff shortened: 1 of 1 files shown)


linux/debian/postinst
```diff
@@ -1,6 +1,17 @@
...
+    if [ $? -ne 0 ]; then
+        echo "ckb-next-daemon unit file not found"
+        echo "restart cannot proceed, please check your install is complete."
+        exit 0;
+    fi
+    systemctl start ckb-next-daemon.service
+elif [ -f /usr/sbin/service ]; then
+    service ckb-next-daemon restart
+else
+    echo "no systemctl binary, no service binary, CKB-Next does not support SysV init"
```
Ravenslofty, line 15: [That's news to me.](https://github.com/ckb-next/ckb-next/blob/master/linux/sysvinit/ckb-next-daemon.in)

PR #846, opened 2022-06-24
Add GitHub Actions to test building and packaging

hummeltech:
This adds automated building and packaging of `ckb-next` for the following Linux distibutions using GitHub Actions:
  - `almalinux:8`
  - `amazonlinux:2`
  - `archlinux:latest`
  - `debian:10`
  - `debian:11`
  - `fedora:35`
  - `fedora:36`
  - `rockylinux:8`
  - `ubuntu:20.04`
  - `ubuntu:22.04`

This should help in discovering build errors for those distributions.

I didn't see any unit tests or anything to run, so it runs `ckb-next-daemon --help` as at least very basic test.

Here's an example of what a workflow run looks like:
https://github.com/hummeltech/ckb-next/actions/runs/2543736710

### DIFF
--- a/.github/actions/cmake/build/action.yml
+++ b/.github/actions/cmake/build/action.yml
@@ -1,0 +1,27 @@
+---
+runs:
+  using: composite
+  steps:
+    - name: Create `cmake` symbolic link
+      run: |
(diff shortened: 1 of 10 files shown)


tatokis:
I don't have much experience, but LGTM. We'll see how it goes.
Thanks!

Issue #868, opened 2022-08-04
K55 PRO XT: WinLock Led Indicator do not work


hlechner:
### Source
Installed using AUR ([ckb-next-git](https://aur.archlinux.org/packages/ckb-next-git))

## Logs & crash reports

Attach:
[ckb-next-report-1659640667.gz](https://github.com/ckb-next/ckb-next/files/9262698/ckb-next-report-1659640667.gz)

```
$ zcat ckb-next-report-1659640667.gz | grep Failed | awk ' BEGIN { first = 8; last = 20 }
{ for (i = first; i < last; i++) { printf("%s ", $i) } print $last }' | sort | uniq
```

result:
```
(bragi_common.c:13): Failed to get property 0x15. Device said it's not supported.  
(bragi_common.c:13): Failed to get property 0x16. Device said it's not supported.  
(bragi_common.c:13): Failed to get property 0x96. Device said it's not supported.  
(bragi_common.c:30): Failed to set property 0x2. Device said it's not supported.  
```

## What is the program's current behaviour?
When **WinLock button** is pressed it does work by disabling/enabling the windows key, however the non-RGB led indicator do not light when the WinLock is enabled.

## What is the expected behaviour of the program?
When the WinLock is enabled, the non-RGB led indicator should light, as it works in hw mode.
![image](https://user-images.githubusercontent.com/11244067/183394322-b5b9912f-9a1a-4883-8f9c-87077aea4b85.png)

## Any other notes & comments?
It do not light the non-RGB led indicator but it does shows on the application (but as a button color):
![image](https://user-images.githubusercontent.com/11244067/182937563-8da01e35-47b1-42c2-9577-511d57032dca.png)
![image](https://user-images.githubusercontent.com/11244067/182937599-c5b9c9c4-2b50-41d0-852a-255b6e72a324.png)

tatokis:
What if you change this
https://github.com/ckb-next/ckb-next/blob/6c5744262a194435ae0f386c430ed780522d1243/src/daemon/keymap.c#L420
to
```
{ "lock",            114, KEY_CORSAIR },
```

hlechner:
> What if you change this
> 
> https://github.com/ckb-next/ckb-next/blob/6c5744262a194435ae0f386c430ed780522d1243/src/daemon/keymap.c#L420
> 
> 
> to
> ```
> { "lock",            114, KEY_CORSAIR },
> ```

[...]

PR #869, opened 2022-08-08
disable led on some keys on K57 and K55 PRO XT

hlechner:
K55 PRO XT and also K57 do not have LED on the following keys:
mr, light, lock, mute, voldn, volup, stop, prev, play, next

Tested on K55 PRO XT

Result:
image

Find it out while trying to find the reason for #868 (not fixed on this PR)


hlechner:
Added commit to hide indicators unsupported by K57_WL on performance tab:
image

There is an extra space that I couldn't hide using the same `->hide();`, so, I kept the line commented.

```
src/gui/kperfwidget.cpp:210:35: error: ‘class QSpacerItem’ has no member named ‘hide’
```

PR #878, opened 2022-09-07
Added functionality for K100 Mechanical

moggesmith10:
Could someone check that this still works on optical keyboards and that I haven't borked anything?


tatokis (changes requested):
Since I believe both models have an identical keymap, they should share the same one in the GUI and so the changes in the GUI (`src/gui`) are unnecessary, and should thus be reverted. A single "K100" keymap seems sufficient.

Everything else seems okay at a cursory glance.

zDEFz:
the back of my keyboard reads "K100 RGB Optical Mechanical Gaming Keyboard"
If there is anything to check, what would it be? All models seem to be optical of that series

zDEFz:
@moggesmith10 git cloned your repo and the UI broke. https://transfer.sh/0Buudt/screenshot_2022-09-26_02-05-43_129282930.png

tatokis:
3ec6f6733ee0002011718a21f0765ab911699199

Issue #864, opened 2022-07-30
Animation settings window has a hardcoded size


tatokis:
Translations or fonts that take more space end up getting cut off. Ideally, the window needs to be made resizeable.

john-yian:
Hello, I am interested in possibly working on this issue if it is still available as a first issue. Thanks!

tatokis:
It is, yes

PR #888, opened 2022-10-02
Allow adjustable window size

john-yian:
Added adjustSize() to showWindow() function to implement adjustable size for issue #864


tatokis:
@john-yian Apologies, perhaps it wasn't clear in the issue (as it was only mentioned in the title).
The issue is about the animation settings dialog having a fixed size and cutting off text.

For example, on my system the text for "Background", "Normalisation", "Vertical Offset", "Audio Device" and "Start with key press" are cut off, and there is no way to resize the window either. 
animsettings

The window should be user-resizable (and should behave correctly when resized), and the default size should avoid cutting off text or controls.

PR #916, opened 2022-12-27
Create daemon mode

felipealmeida:
Initialize in ::main when in daemon mode. Avoid creating window and use QCoreApplication to avoid connecting to display server.

This allows the application to be ran as a service.

You can use the following to install it like this:
https://github.com/felipealmeida/aur-ckb-next-git

### DIFF
--- a/src/gui/kblight.cpp
+++ b/src/gui/kblight.cpp
@@ -472,7 +472,7 @@ void KbLight::load(CkbSettingsBase& settings){
         }
     }
     emit didLoad();
-    map(currentMap);
+    map(currentMap.count() == 0 ? _map : currentMap);
 }
 
 void KbLight::save(CkbSettingsBase& settings){
--- a/src/gui/main.cpp
+++ b/src/gui/main.cpp
@@ -259,57 +259,14 @@ QSettings::setDefaultFormat(CkbSettings::Format);
     if(tmpSettings->value("Program/HiDPI", false).toBool())
         QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
 #endif
-
-    // Setup main application
-    std::unique_ptr<QCoreApplication> app
-      (daemon ? new QCoreApplication(argc, argv) : new QApplication(argc, argv));
-    QCoreApplication& a = *app;
-
-    // Setup translations
-    QTranslator translator;
-    if(translator.load(QLocale(), "", "", ":/translations"))
-        a.installTranslator(&translator);
-
-    const quint16 currentSettingsVersion = tmpSettings->value("Program/SettingsVersion", 0).toInt();
-    if(currentSettingsVersion && currentSettingsVersion > CKB_NEXT_SETTINGS_VER){
-        if(QMessageBox::warning(nullptr, QObject::tr("Downgrade Warning"),
-                                QObject::tr("Downgrading ckb-next will lead to profile data loss. "
-                                            "It is recommended to click Cancel and update to the latest version.<br><br>"
-                                            "If you wish to continue, back up the settings file located at<blockquote>%1</blockquote>and click OK.").arg(tmpSettings->fileName()),
-                                QMessageBox::Ok, QMessageBox::Cancel)
-                == QMessageBox::Cancel){
-            delete tmpSettings;
-            tmpSettings = nullptr;
-            return 0;
-        }
-        // Only handle the downgrade here. The upgrade is handled in CkbSettings.
-        tmpSettings->setValue("Program/SettingsVersion", CKB_NEXT_SETTINGS_VER);
-    }
-
-    // Setup names and versions
-    QCoreApplication::setOrganizationName("ckb-next");
-    QCoreApplication::setApplicationVersion(CKB_NEXT_VERSION_STR);
-    QCoreApplication::setApplicationName("ckb-next");
-
     // Setup argument parser
     QCommandLineParser parser;
     QString errorMessage;
-    parser.setApplicationDescription("Open Source Corsair Input Device Driver for Linux and OSX.");
     bool background = false, daemon = false;
+    {
+      QCoreApplication parser_app(argc, argv);
+    parser.setApplicationDescription("Open Source Corsair Input Device Driver for Linux and OSX.");
 
-    // Although the daemon runs as root, the GUI needn't and shouldn't be, as it has the potential to corrupt settings data.
-    if(getuid() == 0){
-        printf("The ckb-next GUI should not be run as root.\n");
-        return 0;
-    }
-
-    // Seed the RNG for UsbIds
-    Q_SRAND(QDateTime::currentMSecsSinceEpoch());
-
-#if QT_VERSION >= QT_VERSION_CHECK(5, 3, 0)
-    // Enable HiDPI support
-    qApp->setAttribute(Qt::AA_UseHighDpiPixmaps);
-#endif
 
     // Parse arguments
     switch (parseCommandLine(parser, &errorMessage)) {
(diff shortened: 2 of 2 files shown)


Ravenslofty:
So, here's my question: what does this achieve that `ckb-next -b` doesn't. It doesn't need a display server, sure, but if you don't have a display server you can't really configure the GUI anyway.

felipealmeida:
I can copy my .config file to a ckbnext user that runs the program through a systemctl service. So I don't really have to care if the application is running or not. This also means that as soon as the computer boots the keyboard is configured.

felipealmeida:
I've been using this patch for almost two years now.

src/gui/main.cpp
```diff
@@ -65,6 +67,9 @@ CommandLineParseResults parseCommandLine(QCommandLineParser &parser, QString *er
     const QCommandLineOption backgroundOption(QStringList() << "b" << "background",
                                               QObject::tr("Starts in background, without displaying the main window."));
     parser.addOption(backgroundOption);
+    const QCommandLineOption daemonOption(QStringList() << "x" << "daemon",
+                                              QObject::tr("Starts as daemon without display connection."));
```
tatokis, line 71: Please make it obvious that this use case is not officially supported. Fixes will always be accepted of course.

src/gui/main.cpp
```diff
@@ -249,55 +259,14 @@ QSettings::setDefaultFormat(CkbSettings::Format);
...
-    QCoreApplication::setOrganizationName("ckb-next");
-    QCoreApplication::setApplicationVersion(CKB_NEXT_VERSION_STR);
-    QCoreApplication::setApplicationName("ckb-next");
-
     // Setup argument parser
     QCommandLineParser parser;
     QString errorMessage;
+    bool background = false, daemon = false;
+    {
+      QCoreApplication parser_app(argc, argv);
```
tatokis, line 267: 1. You need to call all those static functions before.
```
$ ./bin/ckb-next --version
ckb-next 
```
2. Maybe it would be a good idea to make a QCoreApplication at the beginning, and then _only_ if a GUI is requested, destroy it and make a QApplication.
3. I'm lax about coding style because the codebase is spaghetti. The only thing I require is 4 space indentation, and following in general whatever style is used around in that area. I will not be pointing out every 2 space indentation you used.

src/gui/main.cpp
```diff
@@ -249,55 +259,14 @@ QSettings::setDefaultFormat(CkbSettings::Format);
...
     parser.setApplicationDescription("Open Source Corsair Input Device Driver for Linux and OSX.");
```
tatokis, line 268: If you want, please rebase it against the latest changes, which add CKB_NEXT_DESCRIPTION here.

src/gui/main.cpp
```diff
@@ -352,8 +321,59 @@ QSettings::setDefaultFormat(CkbSettings::Format);
         // If launched with --background, launch in background
         background = true;
```
tatokis, line 322: Maybe move the daemon above, and just make it a // fallthrough.

src/gui/main.cpp
```diff
@@ -352,8 +321,59 @@ QSettings::setDefaultFormat(CkbSettings::Format);
...
+        background = true;
+        daemon = true;
+        break;
+    }
     }
 
+    // Setup main application
+    std::unique_ptr<QCoreApplication> app
+      (daemon ? new QCoreApplication(argc, argv) : new QApplication(argc, argv));
+    QCoreApplication& a = *app;
```
tatokis, line 335: There are very few uses of a. I think it'd be best to just use app directly.

src/gui/main.cpp
```diff
@@ -385,25 +405,35 @@ QSettings::setDefaultFormat(CkbSettings::Format);
     if(background)
         shm_str = nullptr;
 
-    if(isRunning(shm_str) && !QtCreator){
+    if(!background && isRunning(shm_str) && !QtCreator){
```
tatokis, line 408: Won't this allow for multiple instances fighting each other under the same user (possibly messing up the config too)?

src/gui/main.cpp
```diff
@@ -385,25 +405,35 @@ QSettings::setDefaultFormat(CkbSettings::Format);
...
 
-    MainWindow w(silent);
-    if(!background)
-        w.show();
+    std::unique_ptr<MainWindow> w;
+    if (daemon)
+    {
+      AnimScript::scan();
+      KbManager::init(CKB_NEXT_VERSION_STR);
+      Kb::frameRate (30);
```
tatokis, line 421: Does this force 30FPS?

src/gui/main.cpp
```diff
@@ -385,25 +405,35 @@ QSettings::setDefaultFormat(CkbSettings::Format);
...
+      w.reset (new MainWindow(silent));
+      if(!background)
+        w->show();
 
 #ifndef QT_NO_DEBUG
-    if(kwdebug){
+      if(kwdebug){
         KeyWidgetDebugger* d = new KeyWidgetDebugger;
         d->show();
         QObject::connect(&w, &MainWindow::destroyed, [d](){delete d;});
```
tatokis, line 433: w.get()...

```
/home/tasos/ckb-next/ckb-next-felipealmeida/src/gui/main.cpp: In function ‘int main(int, char**)’:
/home/tasos/ckb-next/ckb-next-felipealmeida/src/gui/main.cpp:433:25: error: no matching function for call to ‘QObject::connect(std::unique_ptr<MainWindow>*, void (QObject::*)(QObject*), main(int, char**)::<lambda()>)’
  433 |         QObject::connect(&w, &MainWindow::destroyed, [d](){delete d;});
      |         ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qiodevice.h:45,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qdatastream.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtGui/qregion.h:49,
                 from /usr/include/x86_64-linux-gnu/qt5/QtGui/qevent.h:45,
                 from /usr/include/x86_64-linux-gnu/qt5/QtGui/QCloseEvent:1,
                 from /home/tasos/ckb-next/ckb-next-felipealmeida/src/gui/mainwindow.h:4,
                 from /home/tasos/ckb-next/ckb-next-felipealmeida/src/gui/main.cpp:1:
```

src/gui/main.cpp
```diff
@@ -352,8 +321,59 @@ QSettings::setDefaultFormat(CkbSettings::Format);
...
+    QCoreApplication& a = *app;
+
+    // Setup translations
+    QTranslator translator;
+    if(translator.load(QLocale(), "", "", ":/translations"))
+        a.installTranslator(&translator);
+
+    const quint16 currentSettingsVersion = tmpSettings->value("Program/SettingsVersion", 0).toInt();
+    if(currentSettingsVersion && currentSettingsVersion > CKB_NEXT_SETTINGS_VER){
+        if(QMessageBox::warning(nullptr, QObject::tr("Downgrade Warning"),
```
tatokis, line 344: This should straight up print an error and quit if there's going to be a downgrade in daemon mode.

src/gui/main.cpp
```diff
@@ -1,4 +1,5 @@
 #include "mainwindow.h"
+#include "kbmanager.h"
 #include <ckbnextconfig.h>
 #include <QApplication>
 #include <QDateTime>
```
tatokis, line 5: `#include <memory>`

tatokis (changes requested):
I personally don't want to support this use case, but I do understand that some people want this feature, which is why I've been trying to review/merge this PR.

Being brutally honest here, I considered simply re-doing this myself, as the first version of this PR was completely broken. It is now a bit better after the last round of fixes, in that it doesn't result in the normal use case ending with an immediate abort.

Please, in the future, make sure the thing works and is upstreamable. It's exhausting having to constantly review obviously broken things, as I was in the process of doing before your last push.

If you want to address the feedback, feel free. Otherwise let me know and I'll pick it up from here. Even if the feedback is addressed, I might still end up reordering a few things later on and adding other things, as I simply do not have any more energy for more review rounds.

src/gui/kblight.cpp
```diff
@@ -472,7 +472,7 @@ void KbLight::load(CkbSettingsBase& settings){
         }
     }
     emit didLoad();
-    map(currentMap);
+    map(currentMap.count() == 0 ? _map : currentMap);
```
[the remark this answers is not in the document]
owendelong: Just guessing here, but perhaps there's a problem where if currentMap.count returns zero, map(currentMap) causes a crash, so map(_map) is the safe thing to do.

Likely this scenario isn't possible while the GUI is running, but it might be possible when run as a service prior to login and/or run by a user without a configured profile as yet.

src/gui/main.cpp
owendelong: What is your resistance to supporting this use case? I often run without X on my systems and want my keyboard to reflect my profiles at login. It's especially useful where I'm connected to a KVM switch and color-code the keys according to which system is selected on the KVM.

felipealmeida:
Feel free to redo it.

It actually looks like this patch has some problems when it was rebased and brought things I actually didn't patch it.

I'm just rebasing and working around the changes that have been happening upstream which do change the patch quite a lot. My latest local patch is quite different now because it conflicted with a lot of changes that happened upstream.

But I think it is best to leave for you to implement it correctly rather than use my workaround to base it.

Issue #920, opened 2023-01-10
Support for pairing 


JoePfeiffer:
### Enhancement Details

My Dark Core RGB Pro SE has lost its pairing with its dongle, and there does not appear to be a way to reestablish it without booting into Windows and using their iCue software.  Being able to bring it back through ckb-next would be very much appreciated (or if I simply missed how to do it, instructions would also be appreciated!).

### Additional Details

Add any additional details here.

S-Traut:
After fiixng a re-pairing issue on my Corsair mouse with ckb-next, I might take a look at implementing a solution for pairing devices. However i'm not sure how this can be handle in a "generic" way, because my pairing method may not work with other devices than the IRONCLAW WIRELESS RGB.

JoePfeiffer:
Are you saying you were able to re-pair through ckb-next?  If so, how?  I'm happy to test anything you come up with using my mouse.

S-Traut:
Yes by editing some code and using bragi functions, the goal is to rewrite the pairid and the encryption key of both devices. I've done it manually just for fixing my own mouse right now, but I might take a look at implementing something that handle most of the bragi protocol devices.

S-Traut:
@JoePfeiffer FYI, if you really want to pair through ckb-next, you can grab my branch and build it yourself
#952

RokeJulianLockhart:
> https://github.com/ckb-next/ckb-next/issues/920#event-12676805328

@gravityfargo, what commit or PR completed this?

gravityfargo:
![image](https://github.com/ckb-next/ckb-next/assets/61644499/145d94fc-a6f0-4f48-878a-46a2ab20bdc4)

PR #952, opened 2023-04-10
Feature/device pairing

S-Traut:
Adding a new command to be able to pair bragi devices.
This as been tested with a **CORSAIR IRONCLAW WIRELESS USB MOUSE**

Usage:
```
echo pair ckbX,ckbX > /dev/input/ckbX/cmd
Where left side is the devices you want to pair (mouse/keyboard/...) and right side is the dongle device
```

Workflow used to pair the devices:
```
Generates a new PAIRING_ID (8B)
Generates a new ENCRYPTION_KEY (16B)
Set on the dongle device the pairing id and the encryption key with 2 bragi_write_to_handle executions on 0x5 and 0x6 
(the encryption key must be set exactly after the pairing id otherwise it will no work.)
Set each mouse/keyboard devices precised in parameters the same Pairing ID and Encryption key in the exact same way.
```

Note that this is only a command that can be used through CLI command. There is no GUI integration yet.

The encryption key generation as been done through the /dev/urandom in ``src/daemon/utils.c``


src/daemon/usb.h
S-Traut, line 1: @tatokis What should we do about this?

S-Traut:
link to #920

src/daemon/CMakeLists.txt
```diff
@@ -119,6 +119,8 @@ if (MACOS OR LINUX)
               usb_bragi.h
               bragi_common.h
               bragi_notification.h
+              utils.h
+              utils.c
```
tatokis, line 123: Please move `utils.c` up along with the rest of the .c files.

src/daemon/command.c
```diff
@@ -136,7 +137,7 @@ int readcmd(usbdevice* kb, char* line){
             continue;
         }
         // Reject anything not related to fwupdate if device has a bricked FW
-        if(NEEDS_FW_UPDATE(kb) && command != FWUPDATE && command != NOTIFYON && command != NOTIFYOFF && command != RESET)
+        if(NEEDS_FW_UPDATE(kb) && command != FWUPDATE && command != NOTIFYON && command != NOTIFYOFF && command != RESET && command != PAIR)
```
tatokis, line 140: As per our discussion, this isn't needed, as we can't pair devices in bootloader mode anyway.

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
     // Since we use usbrecv for everything, the devices tell us when they are ready to handle another packet.
     return;
 }
+
+static void bragi_generate_encryption_key(uchar* encryption_key) {
```
tatokis, line 287: 4 space indentation please.

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+  ckb_generate_random(16, encryption_key);
+  for(int i = 0; i < 16; i++) {
```
tatokis, line 289: This should be removed. We did identify that iCUE sends bytes > 127.

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+
+  uchar pairing_id[8];
+  ckb_generate_random(8, pairing_id);
+  ckb_info("Generated new pairing id");
+
+  // Set the dongle device pairing_id and encryption_key
+  bragi_set_device_keys(kb, pairing_id, encryption_key);    
+
+  char* devices = strdup(to);
+  char *token_ptr;
```
tatokis, line 328: `char* token_ptr`

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+  char* device_name = strtok_r(devices, ",", &token_ptr);
+  while(device_name != NULL) {
+    
+    int ckb_id = 0;
+    if(!sscanf(device_name, "ckb%d", &ckb_id)) {
+      ckb_err("Failed to parse device name: %s", device_name);
+      return;
+    }
+
+    usbdevice* device = keyboard + ckb_id;
```
tatokis, line 338: Please lock the dmutex before accessing any part of `device`.

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+    
+    int ckb_id = 0;
+    if(!sscanf(device_name, "ckb%d", &ckb_id)) {
+      ckb_err("Failed to parse device name: %s", device_name);
+      return;
+    }
+
+    usbdevice* device = keyboard + ckb_id; 
+    if(device->status != DEV_STATUS_CONNECTED) {
+      ckb_err("Device not found or connected");
```
tatokis, line 340: Maybe specify which device was not found (probably by ckb%d).

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+  bragi_close_handle(kb, BRAGI_GENERIC_HANDLE);
+}
+
+void cmd_pair_bragi(usbdevice* kb, usbmode* mode, int dummy1, int dummy2, const char* to) {
+  (void)mode;
+  (void)dummy1;
+  (void)dummy2;
+
+  uchar encryption_key[16];
+  bragi_generate_encryption_key(encryption_key);
```
tatokis, line 317: Should probably check to see if the generation succeeded.

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+  ckb_info("Generated new encryption key");
+
+  uchar pairing_id[8];
+  ckb_generate_random(8, pairing_id);
```
tatokis, line 321: Likewise

src/daemon/utils.c
```diff
@@ -0,0 +1,15 @@
+#include "utils.h"
+#include "includes.h"
+
+void ckb_generate_random(unsigned int size, unsigned char* data) {
```
tatokis, line 4: Could directly return ssize_t to indicate success/failure (and also perhaps alert in case of short reads).

src/daemon/utils.c
```diff
@@ -0,0 +1,15 @@
...
+  int rand = open("/dev/urandom", O_RDONLY);
```
tatokis, line 5: I think I'd prefer /dev/random

src/daemon/utils.c
```diff
@@ -0,0 +1,15 @@
...
+  if (rand < 0)
+    ckb_err("Error loading /dev/urandom");
+    
+  ssize_t result = read(rand, data, sizeof(unsigned char) * size);
+
+  if (result < 0)
+    ckb_err("Error generating encryption key");
```
tatokis, line 12: "Generating random data", as this is used for pairing ID too

src/daemon/utils.c
tatokis, line 4: Also, I think size_t is more appropriate for the size parameter. Then it can be called as such: ckb_generate_random(sizeof(whatever_buffer), whatever_buffer);

src/daemon/device_bragi.c
```diff
@@ -281,3 +283,69 @@ void bragi_delay(usbdevice* kb, delay_type_t type){
...
+
+  uchar pkt[BRAGI_JUMBO_SIZE] = {0};
+
+  // Write the PAIRING_ID to the device first
+  bragi_open_handle(kb, BRAGI_GENERIC_HANDLE, BRAGI_RES_PAIRINGID);
+  memcpy(pkt + 7, pairing_id, sizeof(uchar) * 8);
+  bragi_write_to_handle(kb, pkt, BRAGI_GENERIC_HANDLE, sizeof(pkt), sizeof(uchar) * 8);
+  bragi_close_handle(kb, BRAGI_GENERIC_HANDLE);
+
+  // The ENCRYPTION_KEY must be writting directly after the PAIRING_ID
```
tatokis, line 304: "be written"

src/daemon/device_bragi.c
```diff
@@ -233,6 +234,7 @@ int start_dongle_bragi(usbdevice* kb, int makeactive){
     // Force back to SW mode
     // FIXME: Does this make a difference?
     bragi_set_property(kb, BRAGI_MODE, BRAGI_MODE_SOFTWARE);
+    kb->active = 1;
```
tatokis, line 237: This should probably be a separate commit (before the one that introduces the pair commands)

tatokis (changes requested):
Please rebase and squash commits as needed. Ideally in my opinion:
1. Mark bragi dongles correctly as active
2. Implement pair command

Although this is just a guideline, I'd like to keep any chances that have even the slightest chance to regress something as separate as possible (so that it can be easily git bisected down the line).

Once the feedback is addressed, I'll test it on my own hardware.

src/daemon/usb.h
tatokis: Please remove this change from this PR. If you want, you can try adding the generic dongle as a separate device (without marking it as a mouse) and see if it still works. If it does, you can open a new PR. Basically look at where the existing dongle define is used, and add a new one in there too.

src/daemon/device_bragi.c
S-Traut: Just adding a ckb_info call, or do you have some specific checks in mind to assure that this succeeded?

src/daemon/device_bragi.c
S-Traut: I'm not very familiar with "dmutex", can you point me out some documentation / tips about it?

src/daemon/utils.c
```diff
@@ -0,0 +1,20 @@
...
+    if (rand < 0) {
+        ckb_err("Error loading /dev/random");
+        return 0;
+    }
+    
+    ssize_t result = read(rand, data, sizeof(unsigned char) * size);
+
+    if (result < 0) {
+        ckb_err("Error generating random data");
+        return result;
```
tatokis, line 15: close() the fd here before returning

tatokis (changes requested):

PR #971, opened 2023-06-18
OpenRC: rework init.d file

negril:
The original version reimplemented `start()` and `stop()` circumvented the added logic in the default ones.

This simplifies the file and allows the user to configure the service in `/etc/conf.d/cbk-next-daemon`.

One example would be:
```sh
# /etc/conf.d/cbk-next-daemon
command_args="--enable-experimental"
```

### DIFF
--- a/linux/openrc/ckb-next-daemon.in
+++ b/linux/openrc/ckb-next-daemon.in
@@ -1,29 +1,10 @@
 #!/sbin/openrc-run
-# Copyright 1999-2015 Gentoo Foundation
+# Copyright 1999-2023 Gentoo Authors
 # Copyright 2017-2018 ckb-next Development Team <ckb-next@googlegroups.com>
 # Distributed under the terms of the GNU General Public License v2
 
-extra_started_commands="reload"
 command="@CMAKE_INSTALL_PREFIX@/bin/ckb-next-daemon"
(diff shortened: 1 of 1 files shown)


stacyharper:
I got this error on Alpine Linux:

````
$ doas rc-service ckb-next-daemon status
/lib/rc/sh/openrc-run.sh: /etc/init.d/ckb-next-daemon: line 10: start_stop_daemon_args+=--background --stdout /var/log/ckb-next-daemon.log --stderr /var/log/ckb-next-daemon.log: not found
 * ckb-next-daemon: error loading /etc/init.d/ckb-next-daemon
```

Anyone got a clue?

negril:
What shell do you use? Almost certain that the `+=` appending isn't supported for you. I didn't really think about anything not using bash. You could just replace it with `=`. Then you can't set `start_stop_daemon_args` in `conf.d` anymore. But I can rework that.

tatokis:
@negril It has been addressed by https://github.com/ckb-next/ckb-next/pull/983

Can you please verify that it still functions as expected on your end?

negril:
@tatokis yes it's fine. I forgot that OpenRC only requires POSIX sh. So I didn't check for that.

PR #983, opened 2023-07-19
Fix ckb-next-daemon OpenRC service file

stacyharper:
fix my problem reported here: https://github.com/ckb-next/ckb-next/pull/971


tatokis:
Thanks, and apologies for this

PR #1011, opened 2023-11-03
Do not write to removed cmdpath

### LINKED ISSUE #1004: ckb-next terminates when keyboard unplugs
### Enhancement Details

Well, not really unplugs, but...  keyboard is connected through a USB-switch, and when I switch away from the machine, the ckb-next program terminates.  When I switch back, I have to start it manually.
If this is a design decision, I can see reasons why.
I would like to suggest having a settings option where the user can select to have ckb-next up an running independent off plugged-in devices.  Default setting can easily be Off for this setting.

Btw.: Great program!

:-) Roald

### Additional Details

Add any additional details here.

JerryMous3:
This fixes crashing with SIGPIPE on device disconnection.
Fixes #1004.

### DIFF
--- a/src/gui/kb.cpp
+++ b/src/gui/kb.cpp
@@ -215,7 +215,7 @@ Kb::~Kb(){
 
     // Kill notification thread and remove node
     activeDevices.remove(this);
-    if(cmd.isOpen() && notifyNumber > 0){
+    if(QFile::exists(cmdpath) && cmd.isOpen() && notifyNumber > 0){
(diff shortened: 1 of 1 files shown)


tatokis:
Thanks for debugging this! Really appreciated, especially since I can not reproduce this myself.

Is there a reason you're not ignoring or handling the SIGPIPE? There's already a signal handler.

Ultimately, this happens because the daemon has closed its end of the pipe, but my concern is that it's possible to close a pipe without deleting it, or the GUI might try to write before the daemon has unlinked the pipe, in which case QFile::exists will still return true.

Issue #842, opened 2022-06-18
MM700 RGB Gaming Mousepad - 1b9b


pakro:
I would kindly ask if there is someone out there who could add support for the MM700 RGB Gaming Mousepad.

The USB ID is **1b1c:1b9b**

The RGB Zones are depicted below. There is a switch to switch some presets and there is a table explaining the different presets. 

![mousepad_rgb_zones](https://user-images.githubusercontent.com/8904139/174431261-f444593a-635a-4ab6-b4ef-9ce8a79bb658.jpg)

Thank you! 
---

Enter the name of your device here and attach the report generated above.

gravityfargo:
Added to the supported devices wiki as "Requested"

prutheus:
What is the current state for the MM700RGB? Will it be supported soon?

CedricB31:
Hello all,

More info about MM700RGB mousepad, i have sniff USB package with my analyzer (Beagle USB Protocol Anlayser). **It implement the bragi protocol**. As the picture shows, there is effectively 3 RGB zones that can be set with a BRAGI_WRITE_DATA command (Frame example:
` 0x08 0x06 0x00 0x09 0x00 0x00 0x00 R_zone1 R_zone2 R_zone3 G_zone1 G_zone2 G_zone3 B_zone1 B_zone2 G_zone3`.



This table below show an extract of my sniffer package at the first connection the RGB color of every zone are 0xFFFFFF (full sniffing in [test.csv](https://github.com/ckb-next/ckb-next/files/13761484/test.csv))  

| HIDReport | Data |  Comment |
| --------- | ---------------------------------------------------------------------------------- | --------- |
| Out | 08 02 11 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... | read VID |
| In  | 00 02 00 1C 1B 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... | 0x1B1C |
| Out | 08 02 12 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... | read PID |
| In  | 00 02 00 9B 1B 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... | 0x1B9B |
| Out | 08 02 03 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... | read BARGI MODE|
[...]

PR #1050, opened 2024-04-06
Add support for mm700

CedricB31:
Hello,
I added support for the MM700 mousepad in my fork that's why i propose this PR to integrate this new device suport in mainline. Actually the only issue/Bug that's i see in this new integration is the 'limitied key selection surface in GUI' (I couldn't find where modify it).
Capture d’écran du 2024-04-07 10-20-54

Below a view of MM700 GUI with these 3 RGB zones according to the [datasheet](https://user-images.githubusercontent.com/8904139/174431261-f444593a-635a-4ab6-b4ef-9ce8a79bb658.jpg)
Capture d’écran du 2024-04-06 23-23-00

There is an request issue to add this MM700 device (Linked issue #842).

Thx
Cedric


tatokis:
I've done some minor cleanups and an important change while preserving authorship in https://github.com/ckb-next/ckb-next/tree/mm700

As per the comment, it's important not to reorder this enum:
https://github.com/ckb-next/ckb-next/blob/04db68d499c32ccd64d7e1258ee5b122b8d1c2cf/src/gui/keymap.h#L51-L70

I'll try to implement support for non rectangular selection surfaces, and then add it on top and merge it. If I end up not doing it, I'll merge it as is. The ST100 and the polaris both suffer from the same issue.

tatokis:
Sorry, it ended up being a massive pain. (see the [gui-rework-keymap](https://github.com/ckb-next/ckb-next/compare/gui-rework-keymap) branch and https://github.com/ckb-next/ckb-next/commit/3b19a11703dd299ccb1947d445fa3f0e172ab807 ). So I'll merge this for now and will try to deal with this again at some point in the future unless someone manages to figure it out.

PR #1057, opened 2024-04-29
Port to Qt 6

MavChtz:
This PR ports ckb-next to Qt 6.

The build system was modified to be Qt agnostic and an option was added to enable Qt 6 support.
C++17 is now required due to Qt 6.
KStatusNotifierItem was updated as it contains an up-to-date libdbusmenu-qt for Qt 6.


a17r:
> Revert commit 294970fd3ad1 ("[kstatusnotifieritem] Replace
  setAssociatedWidget with setAssociatedWindow") as show/hide does not
  work with a QOpenGLWidget parent

Was there any attempt to fix this upstream? Bundling should be avoided when possible.

Issue #1019, opened 2023-11-30
K70 CORE RGB Mechanical Gaming Keyboard


0dataloss:
### What kind of device is it?

Keyboard

### What is the name of the device?

K70 CORE RGB Mechanical Gaming Keyboard

### Link to Corsair's Product Page

https://www.corsair.com/it/en/p/keyboards/ch-910971e-it/k70-core-rgb-mechanical-gaming-keyboard-black-it-ch-910971e-it

### Additional Details

Tried to save config with a windows machine but it will revert back to Chaos once icue on the windows VM goes off

### Generated Report

Drag your generated report here.
[ckb-next-report-1701382270.gz](https://github.com/ckb-next/ckb-next/files/13518167/ckb-next-report-1701382270.gz)

gravityfargo:
Supported as of e78f03e

Ravenslofty:
@gravityfargo can you update the wiki to list this as supported?

PR #1079, opened 2024-07-19
added support for k70 core rgb (de) (uk)

gaxek:
Fixes: #1019 
As far as I can tell all the lights work, so it should.
I didn't really test anything else beside this, so there is still room for improvement.


OrangeReaper:
I've installed this branch on my system and had no issues on my system; thanks to @gaxek.

J-Hayward99:
Hey,

I'm having some trouble with this branch on my machine, it comes up however the media controls (ie icue button, volume knobs left/right/press) aren't showing up and aren't working properly. Not sure if this is within the confines of your branch but I just wanted to note this since it's an issue

J-Hayward99:
Okay looking into the issue: This does seem to cause the media controls to go haywire. I closed the program and it instantly started working as normal.

I will say though, I had the same issue as with OpenRGB (used separately), so I think Corsair can't write embedded software and got the media controls and the light controls non-isolated, so that probably requires its own fix

gravityfargo:
It looks like this was finished separately in  
e78f03e47126d3e7cabc44d3d3689f86322ba234

josephcrowell:
Yes and it's been merged to master but not released to stable. If you install ckb-next-git you'll get a working version.

Issue #1102, opened 2024-10-11
ABNT2 layout K60: Slash/interrogation key doesn't work


marcavis:
### Affected Devices

K60 RGB Pro Low Profile

### Linux Distribution

Ubuntu 24.04

### Installation Method

Ubuntu package manager (synaptic)

### ckb-next version

ckb-next 0.60

### What did you expect?

When the ckb-next-daemon is running, the slash/interrogation key to the left of Right-Shift stops working. Rarely, it is detected as Reduce Volume (same as Fn+F7) but I have no idea how to reproduce that behavior.

### Additional Details

Output of `sudo journalctl -u ckb-next-daemon -b 0`
[journal.txt](https://github.com/user-attachments/files/17335444/journal.txt)
Keyboard image:
![keyboard](https://github.com/user-attachments/assets/f79dceb2-a2a1-468a-a993-28782194b3b0)

### Generated Report

Drag your generated report here.
[ckb-next-dev-detect-report.gz](https://github.com/user-attachments/files/17335442/ckb-next-dev-detect-report.gz)

PR #1104, opened 2024-10-12
Starting implementation of slash key (ABNT layout)

marcavis:
https://github.com/ckb-next/ckb-next/issues/1102
When discussing the issue where my ABNT layout K60 wasn't reading the slash/interrogation key in the IRC, KitsuWhooa directed me through testing the keyboard and we arrived at this edit, which enabled the slash key while the GUI is running.


tatokis:
@marcavis please give it one final test, and if it works I'll merge it.

marcavis:
Running ckb-next-daemon compiled with this change, with the Ckb-next gui open: the slash/interrogation key works
Running just the ckb-next-daemon compiled with this change withough the GUI: slash/interrogation key doesn't work, needs something like the patch we were testing yesterday.

Issue #884, opened 2022-09-25
ckb-next branch k100-v2 unconfigured / missing wheel


zDEFz:
**Click the `Generate Report` button in the `Settings` tab of the ckb-next user interface and attach the generated report to this issue.**

=> Error creating textual authentication agent: Error opening current controlling terminal for the process (`/dev/tty'): No such device or address

### Source
How and where did you get this program, e.g.:
Branches:
  * k100-v2
  * master

## Logs & crash reports
Not possible to gather

## What is the program's current behaviour?
When I use the wheel above ESC and F1, it prints out 
'bcmnopqrstuvwxyz1234567890['
It keeps then spamming one key which has not received keyup... Weird!

Important to note that the affected control is not available to configure! 
https://transfer.sh/CR2HNg/screenshot_2022-09-26_01-42-46_238167483.png

There is no wheel to configure
https://transfer.sh/aBdxVe/screenshot_2022-09-26_01-43-39_255699811.png

Control Wheel Button -> (Unknown)

## What is the expected behaviour of the program?
If not configured, do nothing.

## Any other notes & comments?

Check required #878  Added functionality for K100 Mechanical #878 

ArchLinux x86_x64
Kernel: 5.19.10-arch1-1
Wayland / Sway

gravityfargo:
This issue has been marked as stale due to age and/or lack of activity. Please make sure to upload a report when submitting an issue.

Please feel free to reopen it if you have more to add or if the issue still needs attention. However, it would be best to [open a new issue](https://github.com/ckb-next/ckb-next/issues/new/choose) and provide a newly generated report issue using our updated templates.

mmtmn:
@gravityfargo could you re-open this issue? I'll fix this if no one is up for it... but I'll have to do it on my own free time if that is the case...

tatokis:
Yeah, sorry, this issue should not have been closed

mmtmn:
All good, thanks for re-opening, I might have some time this sunday or the next to look at it if no one gets to it first

mmtmn:
I've added a fix for it over at #1170

Ergwano:
It would be great getting this resolved since as cat owner having the wheel unintentionally triggered is pretty common. A bunch of inputs triggered in some sensitive locations is dangerous and basically makes this software unusable for me.

mmtmn:
[...]

PR #1172, opened 2025-07-01
reject malformed NKRO packets originating from non-keyboard sources

mmtmn:
Added defensive checks to handle_bragi_key_input() to prevent non-keyboard inputs, such as the K100 iCUE control wheel, from being interpreted as valid NKRO key reports. Sanity checks include packet length bounds and heuristic filtering of modifier byte values to avoid garbage input injection.

This prevents unintentional keypress floods like bcmnopqrstuvwxyz1234567890[.

### DIFF
--- a/src/daemon/keymap.c
+++ b/src/daemon/keymap.c
@@ -556,92 +556,57 @@ static const short hid_codes[256] = {
 // Handled by corsair_kbcopy()
 
 static inline void handle_bragi_key_input(unsigned char* kbinput, const unsigned char* urbinput, int length){
-    // Handle the 01 input and 02 media keys
-    // On the K57 WL length is 16, but on the K95P XT and K60 (1b1c:1bad) it is 21
-    if(urbinput[0] == NKRO_KEY_IN && length >= 16){
-        // Skipping the first two bytes, the following 13 bytes can be copied as-is, with an offset
-        // So let's copy them first before we start bodging
+    // Sanity check: first byte must match, length must be valid
+    if (urbinput[0] == NKRO_KEY_IN && length >= 16 && length <= 64) {
+
+        // Extra guard: urbinput[1] (modifiers) should not be suspiciously high
+        // Modifiers are usually in range 0x00–0xFF, but garbage like 0xFF across many packets may signal wheel bugs
+        if ((urbinput[1] & 0x1F) == 0 && length == 64) {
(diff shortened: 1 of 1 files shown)


mmtmn:
this fixed the #884 for me, I'm new to this repository, it took me around 1h on Sunday and another 20 minutes today to wrap my head around it. Since I'm so new to this, please be kind in case it breaks something somewhere else. It does seem to pass successfully the 13 checks but I do not own a bunch of keyboards to make sure this is up to the highest quality assurance standards.

mmtmn:
if you want to keep the comments as is, the only real change is:

        // Extra guard: urbinput[1] (modifiers) should not be suspiciously high
        // So let's copy them first before we start bodging
        // Modifiers are usually in range 0x00–0xFF, but garbage like 0xFF across many packets may signal wheel bugs
        if ((urbinput[1] & 0x1F) == 0 && length == 64) {
            ckb_warn("Suspicious NKRO packet: possibly malformed input from non-keyboard source (urbinput[1] = 0x%02x)", urbinput[1]);
            return;  // Reject suspicious input
        }

Ravenslofty:
I think we would very much like to keep the comments, as they provide useful historical context.

josephcrowell:
If you could revert the comment removals and limit your pr to just this then it would be much easier to merge and it would be preferable to give get credit for this:

```
    // Extra guard: urbinput[1] (modifiers) should not be suspiciously high
    // So let's copy them first before we start bodging
    // Modifiers are usually in range 0x00–0xFF, but garbage like 0xFF across many packets may signal wheel bugs
    if ((urbinput[1] & 0x1F) == 0 && length == 64) {
        ckb_warn("Suspicious NKRO packet: possibly malformed input from non-keyboard source (urbinput[1] = 0x%02x)", urbinput[1]);
        return;  // Reject suspicious input
    }
 ```

mmtmn:
alright, one sec

mmtmn:
@tatokis  does this look good to be merged?

mmtmn:
@tatokis  any updates on this?

mmtmn:
yeah pretty sus open source repository I guess, i've seen other suggest valid PRs and then get pretty badly and unfairly treated, I'm not trusting this anymore, so i'll do what other did and just close it and never use and recommend against using it

PR #1182, opened 2025-07-16
chore: Improve compilation experience

fleischie:
- Resolve QT deprecation warnings, both are guarded by `QT_VERSION >= QT_VERSION_CHECK()`:
  - Qt 6.5 deprecates `QRegularExpression::match()` in favor of `QRegularExpression::matchView()`, and
  - Qt 6.7 deprecates `QCheckBox::stateChanged()` in favor of `QCheckBox::checkStateChanged()`,
- enable compile_commands.json by default.


src/gui/kbwidget.cpp
```diff
@@ -245,7 +252,7 @@ void KbWidget::currentSelectionChanged(const QModelIndex& current, const QModelI
     device->setCurrentMode(mode);
 }
 
-void KbWidget::batteryTrayBox_stateChanged(int state){
+void KbWidget::batteryTrayBox_checkStateChanged(int state){
```
tatokis, line 255: Same as above

tatokis (changes requested):

PR #1185, opened 2025-07-27
Fix wayland interface detection

TheBill2001:
In PR #1142, I made a mistake in understanding how Wayland protocols' version numbers are reported. An issue occurred when Plasma 6.4 was released and [ext_idle_notifier_v1](https://wayland.app/protocols/ext-idle-notify-v1) version 2 support was added (stupid naming scheme 😠). `WaylandUtils::hasInterface` only returned true if the version is exactly match.

This commit now allows checking Wayland interface existence without requiring a version number by specifying `version = 0` in `WaylandUtils::hasInterface`. Ideally, we should also have functions to check version ranges of interfaces, but I don't think that is currently necessary with how little `ckb-next` is using Wayland's specific functions.

### DIFF
--- a/src/gui/wayland/waylandidletimer.cpp
+++ b/src/gui/wayland/waylandidletimer.cpp
@@ -52,7 +52,7 @@ public:
 };
 
 WaylandIdleTimer::WaylandIdleTimer()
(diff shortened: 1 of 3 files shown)


tatokis:
Thank you for figuring this out and fixing it!

PR #1194, opened 2025-08-08
Add support for darkstar wireless rgb mouse

joncppl:
Recently got this mouse and decided to attempt adding support.

As far as I can tell, everything I've tested is working (all buttons, leds).
I'm not sure what else to test.

Only weird thing I came across/had to change is the DPI update. The separate X/Y (`0x21`, `0x22`)updates were not working (both failed the commented out `bragi_check_success` and did not result in actual DPI change). I looked at the packet capture from the window driver and saw it using a single command to update the DPI (`0x20`) so I added a path for that and removed the separate X/Y sliders from the UI.

Additionally, there are some outstanding warnings that I'm not sure how to approach:
```
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x15. Device said it's not supported.
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x16. Device said it's not supported.
```

and with the dongle
```
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x15. Device said it's not supported.
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x16. Device said it's not supported.
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x2. Device said it's not supported.
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x44. Device said it's not supported.
[I] ckb1: Pairing id: 94 88 67 bb 78 71 42 98 
[I] Found new bragi subdevice 1
[I] ckb2: Subkb vendor: 0x1b1c, product: 0x1bb2
[I] ckb2: Adding device with protocol 1
[I] Setup finished for /dev/input/ckb1
[W] bragi_process_notification (bragi_notification.c:36): ckb1: Unknown bragi notification for dev 2
[W] bragi_process_notification (bragi_notification.c:36): ckb1: Unknown bragi notification for dev 2
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x15. Device said it's not supported.
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x16. Device said it's not supported.
[I] Setup finished for /dev/input/ckb2
```

## screencaps of the lighting / binding pages
image
image


tatokis:
Thank you very much for the PR. I will give it a thorough review when I have the time, however something that concerns me is that 0x1dbc dongle is a generic dongle for all wireless keyboards/mice. So at the very least it should be named as such in the code. My main concern is what happens if someone with another non-supported device now runs the daemon with the same dongle.

It will connect to the dongle, then find out it has an unsupported device. I don't remember what happens in that scenario. As a test case, can you test the code with *just* the wireless dongle and no support for the darkstar itself?

The behaviour I can imagine should be that it disconnects the entire dongle if it finds an unsupported device and then leaves it alone. I do vaguely recall sometimes dongles report ghost devices with 0xffff ids (or some other bogus value), so we need to ensure that in such cases the daemon doesn't drop the dongle.

tatokis:
Also,

> Additionally, there are some outstanding warnings that I'm not sure how to approach:

You can ignore those. It is how the daemon probes the device to find out what it supports. It basically asks it for a feature, and if the device says "I don't support it", it switches to another feature or mechanism, or marks it as unsupported.

In your case, BRAGI_RADIO_APP_VER and BRAGI_RADIO_BLD_VER are not supported likely because your device doesn't have a dedicated RF chip for wireless, so there is no radio firmware to get its version.

joncppl:
> Thank you very much for the PR. I will give it a thorough review when I have the time, however something that concerns me is that 0x1dbc dongle is a generic dongle for all wireless keyboards/mice. So at the very least it should be named as such in the code. My main concern is what happens if someone with another non-supported device now runs the daemon with the same dongle.
> 
> It will connect to the dongle, then find out it has an unsupported device. I don't remember what happens in that scenario. As a test case, can you test the code with _just_ the wireless dongle and no support for the darkstar itself?
> 
> The behaviour I can imagine should be that it disconnects the entire dongle if it finds an unsupported device and then leaves it alone. I do vaguely recall sometimes dongles report ghost devices with 0xffff ids (or some other bogus value), so we need to ensure that in such cases the daemon doesn't drop the dongle.

I performed the test that you described, just support for the dongle but without support for the mouse.
https://github.com/ckb-next/ckb-next/commit/732dc50f2d4000c821add578db12b2e2d05a9c3e

however it continues despite not having the device:
```
[I] ckb3: Subkb vendor: 0x1b1c, product: 0x1bb2
[E] bragi_wl_string (bragi_common.c:216): Unknown product 0x1bb2
[I] ckb2: bragi dongle hotplug thread finished
[I] ckb3: Adding device with protocol 0
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x2. Device said it's not supported.
[W] bragi_get_property (bragi_common.c:13): Failed to get property 0x44. Device said it's not supported.
[I] ckb2: Pairing id: 4b f5 60 cd 8b 8b e0 a9
[I] Setup finished for /dev/input/ckb2
[E] process_input_urb (keymap.c:907): Unknown bragi data received in input thread 00 from endpoint 82
```

The last message is frequently repeated, and since the daemon is holding the endpoint, basic mouse functionality that worked before starting the daemon is blocked.

I tried hacking bragi_dongle_probe (where I see the 0xffff case) such that it also skips unknown pids but that has the same behaviour. I also tried propagating an  error back to _setupusb through start_dongle_bragi  and it gets stuck in resetusb loop. 

With an escape hatch, "good" behaviour (disconnecting from the dongle, thereby restoring generic mouse behaviour): 
https://github.com/joncppl/ckb-next/commit/1291dcda1568db57c6473e43bf767c2bad09f3ff

perhaps this is a good starting point for supporting the dongle, 
but more work would/consideration would be needed to support [multiple devices connected on one dongle](https://help.corsair.com/hc/en-us/articles/4403176744717-SLIPSTREAM-Multipoint-FAQ)

joncppl:
I will gladly rebase/update this MR, but I require a follow up on what to do/what would be accepted.

Summarizing my previous comment
- Treating it as a generic dongle and the driver has poor behaviour for unknown devices (loops)
- Patched the dongle code to ignore dongles with an unknown device: https://github.com/joncppl/ckb-next/commit/1291dcda1568db57c6473e43bf767c2bad09f3ff  If this approach is acknowledged, I will include it in this MR.

PR #1199, opened 2025-08-23
gui: Rework autostart logic so that the DE can configure it

tatokis:
This PR removes the ability of the GUI to manage XDG autostart. On installation, the file is simply placed in `/etc/xdg/autostart/` and then it's up to the DE to manage it further. (Disabling it, modifying the Exec line to add env vars, etc.)

This also avoids confusion for users of i3-type WMs which do not implement XDG autostart on their own, as the ckb-next GUI will no longer claim it can autostart.


src/gui/settingswidget.cpp
```diff
@@ -79,17 +78,36 @@ SettingsWidget::SettingsWidget(QWidget *parent) :
...
+            if(!d.exists() || !d.cd(QStringLiteral("autostart")))
+                continue;
+
+            if(d.exists(QStringLiteral("ckb-next.desktop"))) {
+                qDebug() << "Found autostart file in" << d.absolutePath();
+                // Path copy pasted from the old autorun.cpp
+                static const QDir path(QStandardPaths::writableLocation(QStandardPaths::HomeLocation) + "/.config/autostart");
+                // ckb-next wouldn't let you modify the autostart file before, so if the file exists, it's a pretty good guarantee that it's the one we created
+                // If the user modified it and made it read only, then this won't be able to remove it anyway
+                QFile::remove(path.absoluteFilePath(QStringLiteral("ckb-next.autostart.desktop")));
```
fleischie, line 103: I wonder how important it is, that we inspect at runtime and every start of the GUI, whether an autostart file exists? Especially if the idea is to hoist the responsibility for autostarting to the user anyway.

On the other hand I am using a Desktop Environment, that does not automatically autostart desktop files in a location. So having those left over would not make a difference.

tatokis:
> Quick note:
> 
> I am not at a computer right now and won't be until probably Tuesday or definitely Thursday at the latest. Thus I am unable to run the code. My comments so far are only based on what I understood the code is supposed to do.

No worries. I appreciate you taking the time to look into this.

PR #1200, opened 2025-08-24
Add support for new Ironclaw RGB Wireless dongle.

Nihlus:
This PR adds support for a recent revision of the Slipstream dongle for Ironclaw RGB mice. Behaviour appears unchanged, and it's just the product ID that's changed.

Tested on Linux Mint 22.1 with an Ironclaw bought in 2025 in the EU.


tatokis:
Thank you very much for taking the time to contribute. I really appreciate it. Unfortunately, the 0x1bdc is a generic dongle that is now used for all keyboards/mice. Thus, simply adding it might result in unsupported devices no longer functioning.

Unfortunately, someone else opened a PR adding the dongle id, and I have elaborated further there. Obviously I can't merge both requests (assuming my concerns are addressed), so I am unsure how to proceed.

See https://github.com/ckb-next/ckb-next/pull/1194

Nihlus:
I'm fine dropping this PR in favor of the other one - I don't have enough knowledge about the codebase to propose any true solutions if the dongle is generic, unfortunately, and not enough time to get myself up to speed.

PR #1202, opened 2025-08-31
Add LSB-compliant version of the sysvinit ckb-next-daemon.sh

spillner:
ckb-next's current sysvinit script invokes a number of functions beyond those guaranteed by https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/iniscrptfunc.html.  This patch auto-detects whether those enhanced tools (particularly `start-stop-daemon` appear to be available, and if not builds `ckb-next-daemon.sh` from an alternate LSB-compliant template.

There should be no impact to those using an alternate init scheme, or who have `start-stop-daemon` in a standard location.


tatokis:
My only suggestion is to add a cmake flag that lets you pick which version of the file to generate. (For example someone might be building in a container that doesn't use start-stop-daemon but is targeting a system that has it.)

Other than that, I don't see any glaring issues (but I haven't tested it).

tatokis:
Cheers, thanks!

https://github.com/ckb-next/ckb-next/commit/4bf942dba5e73c2778ef797b6b8dd6b0239aca9a

Issue #1203, opened 2025-08-31
sysvinit ckb-next-daemon.sh "reload" command hangs driver instead of reloading the configuration


spillner:
### Affected Devices

STRAFE (non-RGB)

### Linux Distribution

Slackware

### Installation Method

From git

### ckb-next version

0.6.2-70-g02c2f2df

### What did you expect?

The "reload" command should prompt the daemon to reload its configuration files, and possibly to reset the status of its connection with any supported hardware devices.  Instead, it causes the driver to immediately hang, because the `ckb-next-daemon.sh` script is set up to set SIGUSR1 to the running `ckb-next-daemon` process, but that signal is not handled by the daemon code.

### Did you try anything that may have helped?

Immediately upon issuing the `reload` command, the driver hangs, rendering the keyboard unusable.  The keyboard can be reactivated by disconnecting and reconnecting it to the USB port.

### Additional Details

I'm not sure why `linux/sysvinit/ckb-next-daemon.in` is set up to send SIGUSR1 to the daemon process, since there does not appear to be any corresponding handler for SIGUSR1 in the daemon code.  `src/daemon/main.c` sets up handlers only for SIGTERM, SIGINT, SIGQUIT, and SIGUSR2, with SIGUSR2 doing nothing.

### Generated Report

[ckb-next-dev-detect-report.gz](https://github.com/user-attachments/files/22069216/ckb-next-dev-detect-report.gz)

Ravenslofty:
My knowledge of the daemon internals is a bit outdated, but last time I hacked on it, there were no configuration files for the daemon to reload. So... if anything, I think the actual issue is that the sysvinit script has a reload command to begin with.

spillner:
That was my impression too--- looks like it writes a few files under `/var/run/ckb`, but doesn't read anything other than its own PID.  But I don't know enough about the driver or its history to guess whether there used to be a purpose behind the "reload" command, or it was just unthinkingly carried over from some other piece of software.  For example, it might make sense to scan for newly connected devices, but it looks like `usbmain()` already does that continuously.

spillner:
I reviewed the daemon code; you're right that there are no configuration files to reload; the only place it looks for information is the USB subsystem within udev.  It's also clear that the reason for the keyboard crash is that any uncaught signals will kill the driver without giving it a chance to restore the keyboard devices to ordinary HID mode.

[...]

PR #1204, opened 2025-09-01
Reload USB subsystem on SIGHUP or SIGUSR1

spillner:
Addresses #1203 and makes behavior more consistent with other Linux daemons.  Also cleans up the signal handling interface and exit conditions for `usbmain()` a bit to simplify any future extensions (e.g. if you wanted to provide better insulation against abnormal process termination leaving the keyboard in an unusable state).

This commit has SIGHUP/SIGUSR1 support disabled on MacOS simply because I'm not set up to test there, but it's easy to enable (see comment in `exithandler()`) if anyone else wants to check it out.

### DIFF
--- a/src/daemon/main.c
+++ b/src/daemon/main.c
@@ -207,14 +207,9 @@ void exithandler(int type){
     if (type == SIGTERM || type == SIGINT || type == SIGQUIT) {
         usb_shutdown();
         exit(0);
-#ifndef OS_MAC
-    // For now, ignore the "reload" signals on MacOS, as I'm not set up to test on that platform.
-    // However, anyone who wants to try it could add usbkill() to the bottom of mac_exithandler()
-    // and then remove this #ifdef; I think that should make the MacOS behavior equivalent to Linux.
     } else if (type == SIGHUP || type == SIGUSR1) {
         runstate = RELOAD_DAEMON;
         restore_signal_handlers(RESTORE_LAST);
-#endif
     } else {
         // Never expected to be triggered; we somehow set sighandler() on a signal that was not added to this function's logic.
         char *name = get_temp_signal_name(type);


tatokis:
I am not sure what to make of this. It seems to add a lot of complexity (see: state machine for daemon status) for very little gain. If you want to rewrite it to use sigaction, then go ahead, I'd be happy with that. At the very least, this needs to be broken up into multiple standalone commits though.

But I personally don't see the point in re-initializing the devices. Usually when you want to do that, something has gone wrong, and at that point, there might be a stuck thread somewhere and it will likely fail, so you'd want to (force) restart the whole daemon in the first place. If anything, I'd rather you remove the reload from the sysvinit scripts. I don't know why it was there in the first place.

Also, since no one has stepped up to support macOS, you are free to make breaking changes to the code without having to worry about it.

spillner:
>I am not sure what to make of this. It seems to add a lot of complexity (see: state machine for daemon status) for very little gain.

"State machine" seems rather generous; right now it's really just a flag for "daemon reload has been requested but not yet executed."  I set it up as a generic `runstate` variable only to clarify the semantics, and to provide an easy path for any future additions (e.g. if there were ever a need to break the startup or shutdown process into multiple phases).

>If you want to rewrite it to use sigaction, then go ahead, I'd be happy with that. At the very least, this needs to be broken up into multiple standalone commits though.

Most of the changes by LOC are just cleaning up `main.c` so you only need to list the supported signals in a single place (`signals_trapped[]` definition) instead of having case statements all over the place, removing the baked-in assumption that `SIGUSR2` is the only event that will ever trigger a `nullhandler()`, etc.  I agree that at first glance it looks like a surprisingly long diff for what it fixes, but I think it's fairly straightforward, and a maintainability win overall.

I'd like to get rid of `get_signal_name()`/`get_temp_signal_name()`, but it looks like `sigabbrev_np()` isn't particularly portable, and the return values from `strsignal()` are a lot wordier than the case statements I'm replacing.
 
The only unnecessary complexity is the double-`SIGINT` state machine in `ignore_signal()`; I added that purely for debugging convenience, so that if I somehow mishandle the other termination signals you can at least `^C^C` and probably get a clean daemon exit, instead of having to `kill -9` and definitely losing the keyboard.  However, I'm confident in the other handling code and don't object at all if you want to pull this safety backstop out of `ignore_signal()`.

>But I personally don't see the point in re-initializing the devices. Usually when you want to do that, something has gone wrong, and at that point, there might be a stuck thread somewhere and it will likely fail, so you'd want to (force)

As I noted in #1203, I don't know of a use case where this "reload" is actually preferable to just killing and restarting the daemon process.  However, I think the real bug is that the current daemon exits uncleanly in a whole lot of cases (not just when `ckb-next-daemon.sh reload` is issued), and leaves the system in a pretty awkward state (keyboard probably unusable) when it does so.  In other words, re-initializing the USB subsystem may not be useful, but running `exithandler()` in more cases is.  If you're worried about threads failing, that just reinforces the argument--- I can't think of a reason not to at least try to call `exithandler()` on a `SIGPIPE`, `SIGSEGV`, `SIGABRT`, etc.

The one I think is actually most useful to trap is `SIGHUP`, which is the traditional "reload" mechanism for UNIX daemons.  Classic UNIXes sent `SIGHUP` whenever the parent process exited; in Linux and modern BSDs that's been reduced to only the session-controlling process (not necessarily the immediate parent) designated by `setsid()`, unless the child requests different behavior via `prctl()`.  That isn't a problem if `ckb-next-daemon` is run only via `init`/`systemd`/etc., but it could be problematic for users who manually start it from a shell, perhaps because they only occasionally run `ckb-next`, or are testing it out before deciding whether to make it a default-enabled service.  Anyway, I can't recall seeing too many other daemons that didn't at least trap `SIGHUP`, even if it doesn't trigger a "reload" or other specific behavior.  I can't think of a reason why you wouldn't want to do that here.

>restart the whole daemon in the first place. If anything, I'd rather you remove the reload from the sysvinit scripts. I don't know why it was there in the first place.

That's the simplest fix for the surface issue, but I don't think it leaves the daemon in the strongest overall state.

>Also, since no one has stepped up to support macOS, you are free to make breaking changes to the code without >having to worry about it.

Good to know, thanks!

PR #1209, opened 2025-09-07
Update 99-ckb-next-daemon.rules

robertdonat:
udev errors due to missing comma and whitespace between tokens


tatokis:
Looks good! Just a quick question. Where do you see this error?

robertdonat:
> Looks good! Just a quick question. Where do you see this error?

Thanks! Yeah, it fails when verifying udev rules.

`# udevadm verify`

tatokis:
Awesome! Thank you :)

PR #1219, opened 2025-09-27
daemon: Don't interpret Report ID as NKRO

tatokis:
This prevents the daemon from spamming the logs with unknown NKRO byte messages by no longer attempting to interpret the Report ID as NKRO and ignoring keyups for the warning.

@RoninTech could you test this on your legacy device? Please ensure the alphabet, media, and G keys still function after this change. Thanks!

### DIFF
--- a/src/daemon/keymap.c
+++ b/src/daemon/keymap.c
@@ -614,6 +614,7 @@ static inline void handle_bragi_key_input(unsigned char* kbinput, const unsigned
         CLEAR_KEYBIT(kbinput, 104);         // voldn
 
         // We only care about the first byte
+        // FIXME: Why?
         switch(urbinput[1]){
         case 181:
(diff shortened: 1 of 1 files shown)


RoninTech:
> @RoninTech could you test this on your legacy device? Please ensure the alphabet, media, and G keys still function after this change. Thanks!

I absolutely would but I just hit the road for some traveling and that keyboard is back in Canada.

RoninTech:
> @RoninTech could you test this on your legacy device? Please ensure the alphabet, media, and G keys still function after this change. Thanks!

I absolutely would but I just hit the road for some traveling and that keyboard is back in Canada.

tatokis:
No worries. Let me know when you can test it. I'll merge it in the meantime as I don't expect it to break anything (famous last words). I wanted to be 100% sure.

PR #1231, opened 2025-11-19
Fix: Map rmenu key to KEY_MENU instead of KEY_COMPOSE

Sunny-Pirate:
## Description

This PR fixes the incorrect mapping of the Menu key (rmenu) on Corsair keyboards.

Currently, the Menu key is mapped to `KEY_COMPOSE` (127), which is used for multi-key character composition. This makes the Menu key unusable for its intended purpose: opening context menus.

This PR changes the mapping to `KEY_MENU` (139), the correct Linux keycode for the Menu/Application key.

## Problem

When pressing the Menu key on Corsair keyboards (K70, K95, Strafe, etc.):

Menu key on Corsair K70
*The Menu key (located to the right of Fn key) on Corsair K70 Core RGB*

- ❌ The key sends `KEY_COMPOSE` instead of `KEY_MENU`
- ❌ Many desktop environments show a "ban" symbol or do nothing
- ❌ Applications cannot use the Menu key for context menus
- ❌ The key's primary function is completely broken

Ban symbol in GNOME
*Ban symbol shown when pressing Menu key in GNOME Shell (before fix)*

## Solution

Updated two locations in `src/daemon/keymap.c`:

1. Line 96: Standard keymap - `rmenu` now maps to `KEY_MENU`
2. Line 430: Bragi protocol keymap - `rmenu` now maps to `KEY_MENU`

## Testing

Tested on **Corsair K70 CORE RGB Mechanical Gaming Keyboard**:

### Before fix

```bash
$ sudo evtest
# Pressing Menu key:
Event: type 1 (EV_KEY), code 127 (KEY_COMPOSE), value 1
```

Result: Ban symbol, no context menu

### After fix

```bash
$ sudo evtest  
# Pressing Menu key:
Event: type 1 (EV_KEY), code 139 (KEY_MENU), value 1
```

Result: ✅ Context menu opens correctly

### Desktop Environments Tested

- ✅ KDE Plasma - Native Menu key support works
- ✅ GNOME Shell - Works with Shift+Menu (Shift+F10 equivalent)
- ✅ IntelliJ IDEA - Context menu works with Menu key

## Impact

This fix benefits all users with Corsair keyboards that have a dedicated Menu/Application key, including:

- K70 series
- K95 series  
- Strafe series
- Other models with Menu key

## Breaking Changes

None. `KEY_COMPOSE` was the wrong mapping - this PR corrects it to the standard Linux keycode for Menu keys.

Applications that may have worked around the incorrect `KEY_COMPOSE` mapping will need to update their workarounds, but this is the correct behavior.


tatokis:
> Currently, the Menu key is mapped to KEY_COMPOSE (127), which is used for multi-key character composition.

No, it's not used for multi-key character composition.

> This PR changes the mapping to KEY_MENU (139), the correct Linux keycode for the Menu/Application key.

That is not the correct code. If you run evtest with **any** keyboard that has a menu button (including whatever Corsair one you might have without ckb-next), you can confirm that.

> ❌ Many desktop environments show a "ban" symbol or do nothing
> ❌ Applications cannot use the Menu key for context menus

Citation needed.

> ✅ KDE Plasma - Native Menu key support works

I spent time to test it myself. It absolutely doesn't.

> ✅ GNOME Shell - Works with Shift+Menu (Shift+F10 equivalent)

The menu is supposed to open with just Menu, not with Shift.

> Breaking Changes
> None. 

:)

> Applications that may have worked around the incorrect KEY_COMPOSE mapping will need to update their workarounds, but this is the correct behavior.

I really doubt the kernel would make a change that breaks userspace. I am going to need to see a kernel commit for this.

Finally, I tested this on GNOME 49.1, and KEY_COMPOSE works as expected, whereas KEY_MENU doesn't do anything.

PR #1236, opened 2025-12-01
k70 max basic support

jjmcdn:
I decided to try to get this working for my K70 Max starting from the work to by Tasos Sahanidis back in July. I don't know that the result is "correct" but it's now letting me apply colours and animations to my K70 Max so it seems to be an improvement.


jjmcdn:
@drakej since you seem to have a K70 Max as well, can you confirm this works for you?

restorerootfs:
Just got one of these for Christmas. I'll have a look and see if this works

restorerootfs:
~~Can confirm RGB works fine~~ Update: I can't get the Corsair logo, the media keys, or the Fn key to light up

Solidongle:
I have been able to get the Media Keys and Fn key to light up, all I am missing right now are the Corsair Logo and the Windows/Super Key Lock functionality and the outer LEDS for the spacebar.
Mute button also works to display Mute Status and toggles with pressing.

I was able to make these things work by changing the line:

    LED_CASE_K(P_K70_MAX, 116);

That gets added in led_bragi.c with Daemon's changes to:

    LED_CASE_K(P_K70_MAX, 139);
It seems to me that it's just an issue of addressing the right amount of LEDS, increasing this number to 193 like is done for others makes the entire keyboard lose function, so I'm guessing the proper number for every LED to work lies somewhere between 116 and 193.

Sorry if this is a bad way of commenting, I'm very new to using github.

tatokis:
The correct LED count for this device is 142.

In addition, it appears that the spacebar has 3 LEDs instead of just one, and the same applies to the K70 Pro. It would be appreciated if someone could edit the bragi keymap (in keymap.c) to add the two extra RGB lights and then also add them to the GUI (keymap.cpp). You can use `ckb-next --kwdebug` in a debug build to help with that.

Also, please amend the last commit and remove the change in quickinstall.

tatokis:
It also looks like there are two logo lights

ken-noland-ai-guys:
This works on my device K70 Max. Yes, the spacebar is missing two side lights, the logo is still black, and the mute button doesn't appear to be lighting up, but other than that, I'm running fine.

I'm happy to take a swing at editing the bragi keymap and fixing the remaining LEDs if @jjmcdn isn't around anymore.

Issue #1235, opened 2025-11-29
Unable to create macro using equals key


ayukawazero:
### Affected Devices

Scimitar RGB Elite

### Linux Distribution

Mint

### Installation Method

APT

### ckb-next version

ckb-next 0.6.2-95-gbccdec7

### What did you expect?

Creating a macro binding involving the equals key appears to work as expected until hitting the Apply key, at which point the Equals entries in the events list are immediately removed.  Tested with multiple modifier keys with the same result.

### Generated Report

[ckb-next-report-1764539625.gz](https://github.com/user-attachments/files/23836258/ckb-next-report-1764539625.gz)

itzunagi:
@ayukawazero It looks like this is only an issue when using "All keyboards" to record input.  When recording from a supported keyboard(K100 in my case) the input string is recorded as "equal" rather than "equals".  You should be able to manually use the `Edit as string` option to change "equals" -> "equal" in your macro.

itzunagi:
I believe #1240 should fix the issue, but I'm currently in the process of setting up my machine and can't build it locally to test atm.

tatokis:
Resolved by https://github.com/ckb-next/ckb-next/pull/1240

PR #1240, opened 2025-12-23
Fix key mapping for plus and equal keys when using all keyboards

itzunagi:
When creating a macro using "All keyboards" as input rather than ckb-next devices, the macro generated for equals key is incorrectly including "equals" as the string vs "equal" when using ckb-next supported devices which results in an invalid macro.

See #1235


tatokis:
Thanks for the fix. Cheers!

PR #1251, opened 2026-01-08
Add M75 Wireless (1bf1) support - wired

LuPro:
This adds basic support for the M75 Wireless mouse (only in wired mode), based on work from @Betim-Hodza 

Should be fully featured (button remapping, DPI and lights) and includes a new mouse overlay image (+ svg). ~~Unfortunately I kind of failed aligning the middle mouse button areas for remapping neatly to the center.~~ (Fixed now)

A bit unsure how/whether at all to handle the DPI button on the underside of the mouse. Should that also be rebindable in the UI? Can it even or is that hardcoded in firmware?

~~There are also a few extra changes originally from Betim that I've not fully grasped yet and that seem to alter behavior for other mice as well (so potentially regressing), I'll have to investigate this some more, hence draft for now.~~

Not the cleanest git history so may be necessary to squash before merge.

M75 Support is marked as experimental, not sure though if it should stay this way; in wired mode everything seems to be working with no caveats/hacks/workarounds, it's just that wireless mode doesn't work (at least not through the dongle, haven't tested bluetooth). Not sure what the threshold is for experimental vs "normal" support.

Unsure whether this support also makes the fully wired M75 model work. It probably can since they are identical feature-wise, but I'd guess it has a different USB ID, would be great to hear from someone who has that device.


LuPro:
managed to improve the positioning of mouse buttons on the overlay image (and tweaked the overlay image a bit more to look more like the actual mouse's shape):
image

~~A bit unsure why all the CI checks are failing (happened in previous commit as well), there doesn't seem to be any specific error, just a few warnings in code I didn't touch I think?~~ Fixed now.

ChemoCosmo:
Any chance this will be merged?

PR #1258, opened 2026-02-11
M65 RGB Ultra: Add full support with button fix

rylos:
- Add device support (PID: 0x1b9e) with custom protocol
- Implement RGB lighting control (3 zones: Logo, Wheel, DPI)
- Add DPI control with LED synchronization (up to 26000 DPI)
- Fix button release bug: only 64-byte packets update button state
- Short HID packets now handle only movement and wheel

This fixes the race condition where right/middle/forward/back buttons would release prematurely when moving the mouse while holding them.

Based on martinpal/ckb-next fork (3 commits) + button fix


Brandon-Reichman:
Tested on Arch Linux (kernel 6.1.86-1-lts), GNOME 49.4, Wayland

Device: Corsair M65 RGB Ultra (PID 0x1b9e)

Everything seems to work
- Mouse shows up in ckb-next GUI
- Left, right, middle click
- Mouse movement
- Scroll wheel
- Forward/back buttons
- DPI up/down buttons
- Sniper button remappable 
- RGB lighting control
- Button rebinding
- DPI settings

PR #1276, opened 2026-03-07
Allow control wheel rotation to be bound

fleischie:
This PR is separated into two commits:

## Refactor K100 control wheel setup:

- Move whole cluster more left to be more in line with physical design,
- separate control wheel and media clusters in code, and
- replace individual assignments to `Key`'s attributes by full `Key` assignment to improve readability.

## Add scan codes for clockwise/counter-clockwise rotation of the wheel

- Add keys to `daemon/keymap.c`,
- add `Key`s to `gui/keymap.c`, and
- emit new scancodes for clockwise/counterclockwise rotation instead of hard-coded volup/voldown ones.

(I hope this format is still acceptable, I disclaim, that no LLMs were utilized in making this PR.)

### DIFF
--- a/src/daemon/keymap.c
+++ b/src/daemon/keymap.c
@@ -455,8 +455,8 @@ const key keymap_bragi[N_KEYS_BRAGI_PATCH] = {
     { "prev",          126, KEY_PREVIOUSSONG },
     { "mr",            -1,  KEY_CORSAIR },
     { "profswitch",    128, KEY_CORSAIR },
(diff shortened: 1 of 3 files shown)


tatokis:
quick note: Changelog needs to mention that K100 wheel-as-volume is no longer the case, i.e. "K100 wheel is now unbound by default, you will need to bind it manually". Alternatively we can maybe make the gui bind it to volume by default so that people aren't confused that "it doesn't work".

PR #1279, opened 2026-03-10
Add copy button to "ckb-next-daemon not running" prompt

### LINKED ISSUE #1277: gui: Add copy button to "ckb-next-daemon not running" prompt.
### Enhancement Details

It would be a good UX improvement to have copy buttons for the appropriate commands that people can click on to copy to clipboard. It is perhaps not clear currently that the user can highlight and copy the text.

It is likely easily implementable with QClipboard, although the dialog itself might need to be modified.

### Additional Details

Image

fleischie:
Re-Implement ad-hoc `QMessageBox` as a custom `QDialog`, that adds functionality of copying the service-commands into the clipboard. The commands additionally are keyboard- and mouse-interactable.

Also added DE translations as well (only the dialog title and "Copy to clipboard" are new items, the rest were used in the ad-hoc dialog before).

Coded on a sick day.

Closes #1277

### DIFF
--- a/src/gui/mainwindow.cpp
+++ b/src/gui/mainwindow.cpp
@@ -24,6 +24,7 @@ XWindowDetector* windowDetector = nullptr;
 #endif
 
 #include "ckbsystemtrayicon.h"
+#include "daemonwarndialog.h"
 
 
 extern QSharedMemory appShare;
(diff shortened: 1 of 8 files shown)


src/gui/mainwindow.cpp
```diff
@@ -163,25 +164,7 @@ MainWindow::MainWindow(const bool silent, QWidget *parent) :
...
-    "<blockquote><code>sudo systemctl enable ckb-next-daemon</code></blockquote><br>" +
-    tr("If \"Unit ckb-next-daemon.service is masked.\", unmask it first and try again:") +
-    "<blockquote><code>sudo systemctl unmask ckb-next-daemon</code></blockquote>";
-#else
-    QString daemonDialogText = QString(tr("Start and enable it with:")) +
-    "<blockquote><code>sudo launchctl load -w /Library/LaunchDaemons/org.ckb-next.daemon.plist</code></blockquote>";
-#endif
-    dialog.setInformativeText(daemonDialogText);
-    dialog.setIcon(QMessageBox::Critical);
+    DaemonWarnDialog *warnDialog = new DaemonWarnDialog(mainWindow);
```
tatokis, line 167: This seems like a memory leak. Consider [Qt::WA_DeleteOnClose](https://doc.qt.io/qt-6/qt.html#WidgetAttribute-enum) or another solution

tatokis (changes requested):

src/gui/mainwindow.cpp
fleischie: I added `this->setAttribute(Qt::WA_DeleteOnClose)` to the constructor of `DaemonWarnDialog` now.
I also wanted to note, that it's not a memory leak per se, as the resource is eventually released, when the GUI is closed entirely. But admittedly as the dialog is never shown again, it can (and should) be destroyed when closed.

PR #1262, opened 2026-02-13
Implement a configurable Macro Repetition Delay

fleischie:
Replace the fix repetition delay between macros with a configurable one.

This PR:
- Adds the UI to set (and display) the macro repetition delay,
- amends the GUI and daemon to encode, send, receive, parse, etc. this new data as part of the `$macro:` message,
- does some (potentially unnecessary) cleanup commits in-between to simplify development, and
- update the translation file where necessary and possible.

Screenshot for convenient UI demo:

screenshot_2026-02-13_13-58-15

If merged:
Closes #1253 
Closes #1256

PR #1296, opened 2026-04-12
daemon: provide modifiable repeat delay (i.e. rapid fire)

aerusso:
src/daemon/input.c has DELAY_REPEAT_* which capture the rate at which a macro will repeat if its triggering key is pressed.  That code allows for a different initial, and followup rate, but has hardcoded both for 0.5 seconds.

Separately, some games benefit from "rapid fire" macro actions.  Make this repeat rate adjustable by:

1. Remove the hardcoded 0.5 (expressed in ns) as a variable holding a us delay (microsecond).
2. In _cmd_macro, extend the "macro" command to look for a leading `"=%u;%u;"` in the assignment.  Process the first unsigned integer as the initial repeat delay, and the second as the subsequent repeat delay.

I'm VERY new to any kind of c string parsing, so I apologize if this is bad.  In particular, after writing this, I'm thinking that maybe I should have just done

```c
sscanf(assignment + position, "=%u;%u;%n", &(macro.delay_repeat_initial), &(macro.delay_repeat), &field);
```

However, the patch (as posted, rebased on 0.6.2 in Debian) builds and superficially does what I was hoping for.

I did not try to extend the GUI (I don't use it).  First thing would be to make sure the daemon feature is acceptable/works.  Maybe follow-on work can expose it in the GUI?

P.S. I'm actually somewhat surprised no one added this already, given the path was basically charted out (and I myself worked on it, too!).

### DIFF
--- a/src/daemon/input.c
+++ b/src/daemon/input.c
@@ -82,10 +82,6 @@ static pthread_t macro_pt_first() {
 
 // Default macro keystroke delay
 const struct timespec macrodelay = { .tv_nsec = 1000000 };
-// Initial repeat delay
-#define DELAY_REPEAT_INITIAL 500000000
-// Delay for every subsequent repeat
-#define DELAY_REPEAT_CATCHUP 500000000
 
 static inline void clock_microsleep(uint32_t s) {
     const struct timespec ts = {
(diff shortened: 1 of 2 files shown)


fleischie:
Interesting. Have you noticed #1262 ?

aerusso:
Closing in favor of #1262 .  Funny that we both decided to address this at such a similar time.

PR #1298, opened 2026-04-14
Add support for IRONCLAW Wireless Mouse

LucasionGS:
Fixed up the support for the IRONCLAW Wireless edition of the Corsair mouse. It was missing a couple of buttons, namely:
- Profile Up
- Profile Down
- Opt / Sniper key

They are now all working and remappable now.
It was experimental before but I moved it out of experimental as it now works as it should.


PiniponSelvagem:
Ironclaw wireless SE is not showing up for me.
You have the "original" or the second edition with USB-C?

LucasionGS:
> Ironclaw wireless SE is not showing up for me.
> You have the "original" or the second edition with USB-C?

I have the original with Micro USB

jsuvanto:
I didn't see your PR before making my own. Could you care to test if my PR works on your machine? It is slightly smaller change but perhaps there's some value in combining both.
https://github.com/ckb-next/ckb-next/pull/1337

Issue #863, opened 2022-07-30
Allow selection from a range of colors in animations


RedKnight26:
## What is the program's current behaviour?
Each animation (though I'll be specifically talking about ripple type lighting) only lets you select one color for the animation.

## What would you like it to do instead?
I would like animations to be able to randomly select one color from a preset range of colors so that when each key is pressed it will emit a different colored ripple.

## Any other notes & comments?
I have the attention span of a goldfish and when I'm bored I like watching the pretty colors ripple out on each key press :)

tatokis:
Isn't that what the "Randomly select from gradient" option does?
![695b328ab318](https://user-images.githubusercontent.com/6003656/181914851-c39ccdcc-4b4a-47c2-ad2b-c27e3fb37821.png)

RedKnight26:
> 

That works, but I can't pick rainbow for my gradient. Is there a way to do that?

tatokis:
I'm not sure I understand. Why can't you pick a rainbow for the gradient? There should be a rainbow preset to select.

RedKnight26:
> I'm not sure I understand. Why can't you pick a rainbow for the gradient? There should be a rainbow preset to select.

I think it has something to do with my monitor size. When I click on the gradient, everything's incredibly hard to see. This happens semi-frequently due to my 55" 4k TV
![image](https://user-images.githubusercontent.com/36427975/184056206-c5363fb6-5aa9-402a-9c9d-377b291c91c1.png)

RedKnight26:
quick update, I got it! Though I wish the ripples would be a single color and not actual rainbow ripples

RedKnight26:
update, I got it working!

PR #1302, opened 2026-04-22
Update GradientDialog design file

fleischie:
- Convert main layout to a vertical one, as there is only one column,
- resize preset list widget to allow for more visible items,
- make dialog resizable vertically,
- change flow of preset list to be more intuitive (fill the items left to right and wrap at the width, instead of listing into a horizontal scrolling view),
- display scroll bars in preset list only when necessary, and
- minor changes like spacer changes, etc.

This sort of addresses #863 which originally was due to an already fixed HiDPI scaling issue for the gradient dialog. My hope is, that the issue can be closed because it's more obvious that there are multiple color presets and the dialog is more accessible due to it being vertically resizable.

Before:

screenshot_20260422_091244

After:

screenshot_20260422_091121

screenshot_20260422_091143

Side-note: Pastel color PR is #1301

### DIFF
--- a/src/gui/gradientdialog.ui
+++ b/src/gui/gradientdialog.ui
@@ -10,6 +10,12 @@
     <height>440</height>
    </rect>
   </property>
+  <property name="sizePolicy">
+   <sizepolicy hsizetype="Preferred" vsizetype="Expanding">
+    <horstretch>0</horstretch>
+    <verstretch>0</verstretch>
+   </sizepolicy>
+  </property>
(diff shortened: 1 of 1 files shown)


tatokis:
I wasn't a fan of how it was stretching. How about this? I'll push the changes to this branch if you think it's okay.
Screenshot_20260519_201038

fleischie:
@tatokis Yeah, is good! :slightly_smiling_face:

PR #1325, opened 2026-06-11
New Animation Plugin: Plasma

ewhac:
Renders a [plasma](https://en.wikipedia.org/wiki/Plasma_effect)-style color animation on your devices.  Looks nicest on a keyboard or other large surface with many LEDs.  Rather than playing a fixed animation in a loop, `plasma` effectively displays non-repeating patterns of slowly swirling colors.

The core algorithm is taken from the Plasma screensaver from [ReallySlick Screensavers](http://www.reallyslick.com/screensavers/), converted to work in arbitrary space (rather than pixel space), and had various constants tweaked until I thought it looked nice.  So if the spatial coordinates on your LED devices are laid out correctly, the animation should flow across them fairly well.


ewhac:
Well, this is embarrassing...  The LSP inserted that weird `#include` on my behalf, and I never noticed.  Lemme tidy that up...

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+ * @param context - Pointer to device context.
+ */
+void
+ckb_init (ckb_runctx *context)
+{
+    struct timeval  now;
+
+    // Randomize.
+    gettimeofday (&now, NULL);
+    srand (now.tv_usec);
```
fleischie: I see that other animations do `srand(time(NULL))`, which would also change the `<sys/time.h>` dependency to `<time.h>`.

Also the comment should read "Seed the randomizer." or some such.

CMakeLists.txt
```diff
@@ -106,6 +106,7 @@ cmake_dependent_option(WITH_SNAKE    "Build with 'Snake' animation."    ON "WITH
 cmake_dependent_option(WITH_PIPE     "Build with 'Pipe' animation."     ON "WITH_ANIMATIONS;WITH_GUI" OFF)
 
 if (LINUX)
+    cmake_dependent_option(WITH_PLASMA   "Build with 'Plasma' animation."   ON "WITH_ANIMATIONS;WITH_GUI" OFF)
```
fleischie: I don't see a reason that this animation can only be on LINUX systems. I think this is due to the `<sys/time.h>` dependency, which is not necessary as I outlined in the relevant review comment.

src/libs/uthash/include/uthash.h
```diff
@@ -0,0 +1,1136 @@
...
+PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef UTHASH_H
+#define UTHASH_H
+
+#define UTHASH_VERSION 2.3.0
```
fleischie, line 27: I'd rather you would put this into the `src/libs` directory. This then should behave similarly to kissfft, which is used for the music visualizer animation plugin.

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+
+    // Effect parameters
+    CKB_PARAM_DOUBLE ("zoom", "Zoom factor into plasma field:", "", 20.0, 0.1, 50.0);
+    CKB_PARAM_DOUBLE ("focus", "Focus -- \"sharpness\" of plasma edges:", "", 1, 0.5, 50);
+    CKB_PARAM_DOUBLE ("speed", "Animation speed:", "", 10.0, 0.1, 50);
+
+    // Timing/input parameters
+    CKB_KPMODE (CKB_KP_NONE);
+    CKB_TIMEMODE (CKB_TIME_DURATION);
+    CKB_LIVEPARAMS (TRUE);	// FIXME:?
```
fleischie: What is the FIXME:? comment targeted towards?

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+    // Plugin info
+    CKB_NAME ("Plasma");
+    CKB_VERSION ("0.1");
+    CKB_COPYRIGHT ("2025", "ewhac");
+    CKB_LICENSE ("GPLv2");
+    CKB_GUID ("{15DCB997-B6E2-40D2-9E92-2D3A65938E67}");
+    CKB_DESCRIPTION ("A flowing plasma effect.");
+
+    // Effect parameters
+    CKB_PARAM_DOUBLE ("zoom", "Zoom factor into plasma field:", "", 20.0, 0.1, 50.0);
```
fleischie: Playing around with the animation a bit, I would have put the default zoom a bit lower, e.g. `10`.

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+    CKB_PARAM_DOUBLE ("focus", "Focus -- \"sharpness\" of plasma edges:", "", 1, 0.5, 50);
```
fleischie: Can this just be "Sharpness"? (Also with the variable name?)

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+
+/**************************************************************************
+ * #defines
+ */
+#define TWO_PI          (2.0 * M_PI)
+#define NUMCONSTS       (3 * 6)
+
+#define COUNT_OF(ary)   (sizeof (ary) / sizeof (0[ary]))
+
+#define fabstrunc(v)    (v <= 0  ?  (v >= -1.0 ? -v : 1.0)  :  (v <= 1.0 ? v : 1.0))
```
fleischie: Is this necessary? The animations seem to be linked with `-lm` by default and there is a `fabs()` and a `truncf()` function in the math library.

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * @param context - Pointer to device context.
+ * @param delta - "Duration" since last frame.  May be thought of as "progress" through Playback:Duration for looping animations.
+ */
+void
+ckb_time (ckb_runctx *context, double delta)
+{
+    float   *ps = s;
+    float   *pa = a;
+    float   *pda = da;
```
fleischie: I understand you want to align the variables, but why indent the variable name unnecessarily?

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+
+    for (int i = COUNT_OF (s);  --i >= 0;  ++ps, ++pa, ++pda) {
+        // FIXME: Adjust according to delta.
```
fleischie: Do you still need to fix this FIXME?

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+        *pda++ = (float) rand() / (float) RAND_MAX * 0.005;
+    }
+
+    /*  Allocate memory to hold hashmap entries.  */
+    if (!dotmem) {
+        dotmem = malloc (sizeof (*dotmem) * context->keycount);
+        if (!dotmem) {
+            return;
+        }
+    }
```
fleischie, line 209: Shouldn't this be in `ckb_init()`?

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+    for (pd = dotmem, key = context->keys, i = context->keycount;
+         --i >= 0;
+         ++pd, ++key)
+    {
+        pd->key = key;
+
+        pd->r =
+        pd->g =
+        pd->b = 0.0;
+
```
fleischie: To improve legibility, I would explicitly add the `0.0` value to each line.

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+// Memory holding plasmadots.
+struct plasmadot    *dotmem;
+
+
+/**************************************************************************
+ * Code.
+ */
+void
+ckb_info (void)
```
fleischie, line 91: Could you remove all the spaces between the function name and the parenthesis? To be stylistically consistent with the other animation programs.

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+double  g_speed;
+
+//  "Plasma" coordinate space.
+float   g_wide, g_high;
+float   g_long, g_short;
+float   g_aspect_ratio;
+
+float   s[NUMCONSTS];   // sines
+float   a[NUMCONSTS];   // angles: 0 <= a <= TWO_PI
+float   da[NUMCONSTS];  // delta angles
```
fleischie: Best to type out the names, instead of hiding their meaning in the comments.

src/animations/plasma/main.c
```diff
@@ -0,0 +1,334 @@
...
+ * Globals.
+ */
+double  g_zoom;
+double  g_focus;
+double  g_speed;
+
+//  "Plasma" coordinate space.
+float   g_wide, g_high;
+float   g_long, g_short;
+float   g_aspect_ratio;
```
fleischie: The `g_` prefix is not necessary.

fleischie (changes requested):
Thank you for the submission, I rather enjoyed the animation myself.

I have mostly stylistic review. Additionally can you squash all of commits, as they are not providing any value on their own? (It's for keeping the history clean.)

tatokis:
One thing from my end:

In [Delete debugging printf()s](https://github.com/ckb-next/ckb-next/pull/1325/commits/25bfa5360b89f801cee026eee64d17e66c4a8e8f), if any of them provide value for people debugging something in the future, then instead of deleting them use the DBG() macro and leave them in.

src/animations/plasma/main.c
ewhac: I haven't done a performance analysis, but using `fabs()` and `trunc()` would require two function calls; whereas this requires only two in-line comparisons.

src/animations/plasma/main.c
ewhac: [ewhac 2021 C Coding Guidelines](https://ewhac.org/ewhac-c-coding-style/)

src/animations/plasma/main.c
ewhac: Most of the nomenclature is taken from ReallySlick Screensavers.  That said, I'm not wedded to any particular naming here.

src/animations/plasma/main.c
ewhac: It's been a minute since I wrote that.  I think we can drop the comment.

src/animations/plasma/main.c
ewhac: It was unclear how efficient one should be with memory allocations.  Since we free `dotmem` when `ckb_start()` is called with `state == false`, then yeah, the alloc needs to be here.  But if it's okay for `dotmem` to persist even when the plugin is inactive, then sure, we can move it.

src/animations/plasma/main.c
ewhac: It should probably be a TODO rather than a FIXME.  But I can chop it.

src/libs/uthash/include/uthash.h
ewhac: Didn't know that was a place I could drop things.  Will do.

CMakeLists.txt
ewhac: I see no reason, either, but since Linux is my only environment, I wasn't prepared to claim it built/worked anywhere else.

src/animations/plasma/main.c
fleischie: As a user this was a bit confusing, as I was not sure, what "Focus" meant and needed a second expression to convey it's meaning. I don't think everybody knows the ReallySlick screensavers enough to understand immediately. :sweat_smile:

src/animations/plasma/main.c
ewhac: Just tried it -- that looks good, too.  (I did all my development on a Corsair K95 RGB keyboard -- the old one with 18 macro keys -- and a Razer MM800 RGB Polaris mouse mat.)

src/libs/uthash/include/uthash.h
ewhac: ...Actually, you may have to show me how -- my CMake-fu is not strong.

src/animations/plasma/main.c
ewhac: After kluging together a few experiments, it appears compilers have gotten pretty smart about this stuff.  I'll rework this.

ewhac:
> ...can you squash all of commits, as they are not providing any value on their own? (It's for keeping the history clean.)

Assuming you don't have control over that (GitHub offers a squash-merge on PRs, doesn't it?), that will be the last step once all changes are in place and approved.

src/animations/plasma/main.c
ewhac: The latest checkin changes it to "Sharpness".

src/animations/plasma/main.c
ewhac: Prefix dropped.

src/animations/plasma/main.c
ewhac: Arrays renamed.

src/animations/plasma/main.c
ewhac: Done.

src/animations/plasma/main.c
ewhac: Deleted.

src/animations/plasma/main.c
ewhac: There used to be another variable in that list.  Fixed.

src/animations/plasma/main.c
ewhac: Converted it to a function that more clearly explains what it does.

src/animations/plasma/main.c
ewhac: Changed.

ewhac:
Please see latest checkin.

src/animations/plasma/main.c
fleischie: Fair enough.

src/animations/plasma/main.c
fleischie: I guess it's alright for now. I would suggest keeping it as it is, and check for memory performance when the situation requires it.

src/animations/plasma/COPYING
```diff
@@ -0,0 +1,338 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
```
fleischie: `ckb-next` already provide the GPLv2 license on its root. Is it strictly necessary to provide the license file again as part of this animation? I might be wrong, I don't understand the software licensing in its entirety unfortunately.

You already claimed copyright in the file itself and - as long as its part of this project - the user gets an appropriate license file.

src/libs/uthash/include/uthash.h
fleischie: I wrote a basic "guide" for vendoring in the library into ckb-next (by hand I might add). It should be a good starting point: https://gist.github.com/fleischie/91a0c22f37b1b678936835de59271cc9

Let me know if you need additional information .. or join the IRC, I usually lurk there.

CMakeLists.txt
fleischie: I mean officially we don't support any other OS at the moment ... we don't officially support anything for that matter ... but there is nothing that ties it to a specific LINUX-only library.

So, you can move it out of the conditional. When there will be a problem report, we can always move it back.

src/animations/plasma/COPYING
ewhac: Fair enough; I'll chop it.

ewhac: Deleted.

src/libs/uthash/include/uthash.h
ewhac: Moved.

CMakeLists.txt
ewhac: Moved.

src/libs/uthash/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+project(UtHash VERSION 2.3.0)
```
fleischie, line 30: One final touch: Add the same copyright header as the other `src/libs/.../CMakeLists.txt` without mentioning the `<ckb-next@googlegroups.com>` (that doesn't exist anymore).

fleischie (changes requested):

src/libs/uthash/CMakeLists.txt
ewhac: Done.

src/libs/uthash/CMakeLists.txt
```diff
@@ -1,3 +1,32 @@
...
+#   ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+#   LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+#   CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+#   SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+#   INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+#   CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+#   ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#   POSSIBILITY OF SUCH DAMAGE.
+#
+#   See COPYING regarding the distribution of kissfft library and not this file.
```
fleischie: This should probably say "See Copyright notice in uthash.h file regarding its distribution." or some such.

fleischie (changes requested):

src/libs/uthash/CMakeLists.txt
ewhac: Done.

fleischie:
I squash-merged the PR manually. GitHub seems to have problems with it.

Don't panic, I'll close this PR but it's gratefully merged! :pray: 

Edit: For reference: 833ab50951e230674bda02e8448ef6ef365dfd81

PR #1327, opened 2026-06-23
Add Restart Animation button to KbLightWidget

### LINKED ISSUE #335: Add Restart Animations button in the GUI
It would be useful to have a button in the GUI to restart the whole animation sequence in the mode, as it can be useful when building sequenced animations

Cosky00500:
Closes #335

This PR adds a dedicated "Restart" button inside the general lighting panel (KbLightWidget), fulfilling the feature request to have a restart animation button in the GUI.

Currently, to force-restart an animation, users have to manually switch profiles or reload the GUI. This button provides a quick and clean way to restart the active hardware/software animation directly from the lighting settings.

- Added QPushButton for restarting animations in kblightwidget.ui
- Connected the button to the appropriate animation restart slot in kblightwidget.cpp


src/gui/kblightwidget.ui
```diff
@@ -220,6 +220,19 @@
          <string>New animation...</string>
         </property>
        </widget>
+      </item>
+	<item row="1" column="2">
```
fleischie, line 224: Please make this spacing consistent with the surrounding file, i.e. 6 spaces (I think).

Also I personally think, that this button should not be on the same row as the lighting indicator. Hence increase the `row` attribute here and in the `item` element in row 118.

fleischie (changes requested):
Functionally I think this does, what was intended. 👍 
Edit: OK, no I was informed, that this is not sufficient, yet. Will update - or write another - comment, when I understand the intention more.

PR #1337, opened 2026-08-07
Fix Ironclaw RGB Wireless bindings

jsuvanto:
This should fix non-working button bindings (Profile up, Profile down and Opt/Sniper) on Ironclaw RGB Wireless.

This was fixed using AI tools. Tested to work wirelessly (slipstream dongle) and with USB cable attached on Nobara OS 44.

Note: The device appears in the GUI as different device depending whether USB cable is attached or not, so you have to bind the buttons twice! This is probably true for lighting and other settings as well and is outside the scope of this PR.

### DIFF
--- a/src/daemon/keymap_patch.c
+++ b/src/daemon/keymap_patch.c
@@ -28,6 +28,9 @@ static const keypatch m95patch[] = {
 };
 
 static const keypatch icwpatch[] = {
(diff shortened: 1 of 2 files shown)


jsuvanto:
I see https://github.com/ckb-next/ckb-next/pull/1298 implements the same, i did not notice that PR before submitting this one. This is a slightly smaller change.

Issue #505, opened 2019-11-13
K57 RGB Wireless Keyboard


Brainwasher06:
Hello,

Is it possible to add my keyboard in your software ? If you do it would be much appreciated =)

- If possible, try the latest code by building the software manually. ( i used ./quickinstall , dont know if it's what you mean)

Thank you for your work and your time.

Have a good day
---

Enter the name of your device here and attach the report generated above :
K57 RGB Wireless Keyboard
[ckb-next-report-1573647243.gz](https://github.com/ckb-next/ckb-next/files/3841018/ckb-next-report-1573647243.gz)

tatokis:
It uses a completely new and redesigned protocol that we haven't looked into, and probably will not for a while. No one has shown interest in implementing this protocol.

oppositionsecurity:
Just grabbed this keyboard and would be interested as well.

87racer:
Just grabbed this keyboard and would be interested as well.

MarcinOrlowski:
You guys may want to write to Corsair's support pointing this ticket. Maybe they appear smart and provide you all required docs. Win-Win.

kaijuagenda:
Recently bought one myself. shame there's no support yet.

oppositionsecurity:
Recently returned 2 K57's actually. If you type 70+ WPM you will be sorely disappointed with the "selective n-key rollover". Very common 3 and 4 letter combinations will fail to type in rapid succession. As a side note the K95 is great, despite the cost, and is fully supported and allows for some cool effects using the top LEDs as notifications lights.

kaijuagenda:
> Recently returned 2 K57's actually. If you type 70+ WPM you will be sorely disappointed with the "selective n-key rollover". Very common 3 and 4 letter combinations will fail to type in rapid succession. As a side note the K95 is great, despite the cost, and is fully supported and allows for some cool effects using the top LEDs as notifications lights.

Not sure how well the measure holds up. I frequently break 90+ WPM on the K57 with no apparent issues, Nor can i replicate the problem you described. Either way I don't see the point of this response, unless the goal was to shill a more expensive peripheral.

superjcvd:
I am interested too :)
So far the keyboard works fine in bluetooth.
- the sound control keys are working
- the RGB isn't working
- the macro keys are not working

ThisNekoGuy:
I just got mine in the mail today, myself, it feels pretty great Lol
But yeah, it's unfortunate that there isn't support for it yet

[...]

PR #1338, opened 2026-08-07
#505 Implement K57 RGB Wireless

jsuvanto:
This sets the correct endpoints for K57 RGB Wireless when using in dongle mode. AI tools were used.

Tested on Nobara OS 44. Media controls and G1-G6 buttons work as expected.

Very rarely (maybe once per week) wrong key codes are sent to the system. I don't have a fix for that. Could be something in the endpoint traffic that gets parsed wrong. As a workaround, turning the keyboard on and off fixes that. I feel like this is not a big problem since K57 is still in experimental mode. I will continue investigating this on my leisure.


jsuvanto:
There was a breaking bug in the original PR commit. I have fixed it.

PR #1343, opened 2026-08-14
Adding support for Corsair SABRE RGB PRO  USB 1b1c:1b79

DrewGitsIt:
Added input and lighting support for Corsair SABRE RGB PRO (USB 1b1c:1b79).

What changed:
I started by copying the fairly similar Corsair M55 mouse, which is already supported. ~~But, as I found via Wireshark capture, the lighting control scheme violates an assumption that is generally held in ckb-next code for bragi devices: that planar RGB formats will have zones = stride at the input array. The Sabre RGB Pro has 3 illumination zones and 5 lights to paint (logo, wheel, 3x DPI indicators). This change adds custom logic for the DPI indicator because we are choosing which LEDs to illuminate based on DPI triggers, not lighting triggers. And this one aside allows us to reuse the rest of the generic bragi code. I added a block to handle this in [led_bragi.c](https://github.com/ckb-next/ckb-next/pull/1343/changes#diff-c4fd115e887349ea04bc9fb56087b89785757d51db1dbadcc85cfb0e3fb46fb0).~~ I found a simpler and more idiomatic way to handle lighting the DPI LEDs when the user has them controlled by Lighting. It's based on the implementation for the Dark Core RGB Pro.

Aside from this chunk, most of the other changes are bog-standard adding a new product stuff. I will call out one item that may invite conversation [device_bragi.c](https://github.com/ckb-next/ckb-next/pull/1343/changes#diff-c3aff3f79e37e4bd8f8acd39acd6ae079592025bf4bbb71228bed078fc3d05f2). This line breaks the cached DPI value when transitioning between hardware mode (no daemon) and software mode (daemon running) to ensure that the mouse is never unintentionally set to an invalid DPI. This is purely defensive and would have been helpful while I was developing so I figured I'd pass this benefit on :)


kymikoloco:
Thank you @DrewGitsIt - I built and installed this PR and I can now turn off the super bright rainbow demo!! I don't know how to add new animations, and I can't save it to the hardware, but I only wanted the lights off on Linux.

DrewGitsIt:
> Thank you @DrewGitsIt - I built and installed this PR and I can now turn off the super bright rainbow demo!! I don't know how to add new animations, and I can't save it to the hardware, but I only wanted the lights off on Linux.

I'm happy it helped you! Saving the settings to hardware is not supported on any devices of this class (at least via ckb-next), to my knowledge. But you should be able to add animations in the Lighting tab of the ckb-next GUI. Presets are on the left and the "New Animation"  button operates on whichever lighting zones you select.
Screenshot_2026-09-01_18-06-51

edit: in my latest commit, you should also be able to re-map lighting on each DPI indicator if it's not controlled by the colors in the Performance tab.

